PR #3, opened 2020-05-01
[serde reflection] fix interaction between trace_value and trace_type in enums

matbd:
## Summary

`trace_type` was assuming that we find the variants of an enum starting with index 0. However, this doesn't need to be the case if we use trace_value first.

The new code is not making assumptions on the tracing state any longer. In the worst case, it may have to scan indexes to find a new variant, but the normal case is still O(1).

Also improve the documentation while we're at it.

## Test Plan

See new targeted unit test

### DIFF
--- a/serde-reflection/src/de.rs
+++ b/serde-reflection/src/de.rs
@@ -432,29 +432,29 @@ impl<'de, 'a> de::Deserializer<'de> for Deserializer<'de, 'a> {
             .registry
             .entry(name)
             .unify(ContainerFormat::Enum(BTreeMap::new()))?;
-        let current_variants = match self.tracer.registry.get(name) {
+        let known_variants = match self.tracer.registry.get(name) {
             Some(ContainerFormat::Enum(x)) => x,
             _ => unreachable!(),
         };
         // If we have found all the variants OR if the enum is marked as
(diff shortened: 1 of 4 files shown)


serde-reflection/src/de.rs
```diff
@@ -440,26 +436,29 @@ impl<'de, 'a> de::Deserializer<'de> for Deserializer<'de, 'a> {
...
+            index
         };
-        // Temporarily mark the entry as "incomplete" to force variant #0 in any
+        let mut variant = current_variants.get(&index).cloned().unwrap_or(Named {
+            name: variants
+                .get(index as usize)
+                .expect("variant indexes must be a range 0..variants.len()")
+                .to_string(),
+            value: VariantFormat::Unknown,
+        });
```
bmwill, line 460: So if the variant at index 0 is also unknown, we'll still return an "Unknown" value?

matbd: Yes, every time we investigate a variant index for the first time we start with this "Unknown" placeholder. The new code doesn't assume that 0 is always studied first.

matbd: I should probably rename `current_variants` into `known_variants` for clarity.

Issue #5, opened 2020-05-01
[Feature] Ensure termination of trace_type on more recursive types (e.g. trees)


matbd:
# 🚀 Feature Request

## Motivation

The current code contains some termination logics that works for recursive enums but it is not strong enough yet to converge on trees (using a Vec type for children) and lists (using an Option type for the tail -- see tests below).

Currently, whenever we have a comment `// Finally, update the registry` in the code (see `deserialize_newtype_struct`, `deserialize_tuple_struct`, `deserialize_struct`, and `deserialize_enum`), the current code is delaying registry updates that would be in fact very useful for the termination logic in `deserialize_option`, `deserialize_seq`, and `deserialize_map`.

Delaying updates was needed to pass the borrow checker -- which won't let us take mutable references simultaneously on the registry AND on pieces of it, for good reasons.

This feature will likely require RefCell's or some CPS programming with closures to repeatedly update the registry.

## Test plan

```rust
 #[test]
fn test_tracing_deserialization_for_lists() {
    #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
    struct Node {
        data: u64,
        next: Option<Box<Node>>,
    }
    let samples = Samples::new();
    let mut tracer = Tracer::new(TracerConfig::default());
    tracer.trace_type::<Node>(&samples).unwrap();
    tracer.registry().unwrap();
}

#[test]
fn test_tracing_deserialization_for_trees() {
    #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
    struct Node {
        data: u64,
        children: Vec<Node>,
    }
    let samples = Samples::new();
    let mut tracer = Tracer::new(TracerConfig::default());
    tracer.trace_type::<Node>(&samples).unwrap();
    tracer.registry().unwrap();
}
```

matbd:
landed

PR #7, opened 2020-05-03
[serde-reflection] support recursive types

matbd:
## Summary

This PR aims to address #5 by introducing a more powerful unification algorithm with proper variables and term-sharing.

## Test Plan

included unit tests

### DIFF
--- a/serde-reflection/Cargo.toml
+++ b/serde-reflection/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-reflection"
-version = "0.1.0"
+version = "0.2.0"
 description = "Extract representations of Serde data formats"
 documentation = "https://docs.rs/serde-reflection"
 repository = "https://github.com/facebookincubator/serde-reflection"
--- a/serde-reflection/src/de.rs
+++ b/serde-reflection/src/de.rs
@@ -193,14 +193,11 @@ impl<'de, 'a> de::Deserializer<'de> for Deserializer<'de, 'a> {
     where
         V: Visitor<'de>,
     {
+        let mut format = Format::unknown();
         self.format
-            .unify(Format::Option(Box::new(Format::Unknown)))?;
-        let format = match self.format {
-            Format::Option(x) => x,
-            _ => unreachable!(),
-        };
-        if **format == Format::Unknown {
-            let inner = Deserializer::new(self.tracer, self.samples, format.as_mut());
+            .unify(Format::Option(Box::new(format.clone())))?;
+        if format.is_unknown() {
+            let inner = Deserializer::new(self.tracer, self.samples, &mut format);
             visitor.visit_some(inner)
         } else {
             // Cut exploration.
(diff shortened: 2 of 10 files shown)


serde-reflection/src/format.rs
```diff
@@ -261,6 +321,92 @@ where
...
+        Self(Rc::new(RefCell::new(content)))
+    }
+
+    fn borrow(&self) -> Ref<Option<T>> {
+        self.0.as_ref().borrow()
+    }
+
+    fn borrow_mut(&mut self) -> RefMut<Option<T>> {
+        self.0.as_ref().borrow_mut()
+    }
```
bmwill: nit: maybe use the Borrow/BorrowMut traits?
(i'm fine if you do leave this as is)

serde-reflection/src/format.rs
```diff
@@ -85,12 +99,16 @@ pub struct Named<T> {
...
+pub struct Variable<T>(Rc<RefCell<Option<T>>>);
+
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
 #[serde(rename_all = "UPPERCASE")]
 /// Description of a variant in an enum.
 pub enum VariantFormat {
-    /// A variant whose format is yet unknown.
-    Unknown,
+    /// (used internally for tracing) A variant whose format is initially unknown.
+    Variable(Variable<VariantFormat>),
```
bmwill, line 111: If this is for internal use only then you could always #[doc(hidden)] this

serde-reflection/src/de.rs
```diff
@@ -290,12 +278,8 @@ impl<'de, 'a> de::Deserializer<'de> for Deserializer<'de, 'a> {
     where
         V: Visitor<'de>,
     {
-        self.format
-            .unify(Format::Tuple(vec![Format::Unknown; len]))?;
-        let formats = match self.format {
-            Format::Tuple(x) => x,
-            _ => unreachable!(),
-        };
+        let mut formats: Vec<_> = std::iter::repeat_with(Format::unknown).take(len).collect();
```
bmwill, line 281: Clever way to rewrite `vec![]`

serde-reflection/src/format.rs
```diff
@@ -332,11 +341,11 @@ impl<T> Variable<T> {
         Self(Rc::new(RefCell::new(content)))
     }
 
-    fn borrow(&self) -> Ref<Option<T>> {
+    pub fn borrow(&self) -> Ref<Option<T>> {
         self.0.as_ref().borrow()
     }
 
-    fn borrow_mut(&mut self) -> RefMut<Option<T>> {
+    pub fn borrow_mut(&self) -> RefMut<Option<T>> {
```
bmwill, line 348: Why the change on mutability? and i'm expecting that you want these functions to be part of the public API?

serde-reflection/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "serde-reflection"
-version = "0.1.0"
+version = "0.2.0"
```
bmwill, line 3: Good this is a breaking change :)

serde-reflection/src/format.rs
```diff
@@ -261,6 +321,92 @@ where
...
+
+    fn visit_mut(&mut self, f: &mut dyn FnMut(&mut Format) -> Result<()>) -> Result<()> {
+        match self.borrow_mut().deref_mut() {
+            None => Err(Error::UnknownFormat),
+            Some(value) => value.visit_mut(f),
+        }
+    }
+
+    fn unify(&mut self, _other: Variable<T>) -> Result<()> {
+        // TODO
```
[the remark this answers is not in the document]
matbd: Actually, we don't use this function (and cannot easily implement it without making `FormatHolder` bigger) so I removed the TODO in the next diff. Let me add a comment.

serde-reflection/src/format.rs
```diff
@@ -261,6 +321,92 @@ where
...
+    T: Deserialize<'de>,
+{
+    fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        use serde::de::Error;
+        Err(D::Error::custom("Cannot deserialize variables"))
+    }
+}
```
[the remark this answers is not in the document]
matbd: Let me try `deserialize_with`.

serde-reflection/src/format.rs
matbd: I'm thinking a user may still wish to write `Variable(_) => unreachable!()` instead of `_ => unreachable!()` at the end of a pattern matching.

I also thought about having a separate public type but one would need to change trace_type and trace_format APIs (to not return a temporary format) and make unit tests more complicated.

serde-reflection/src/de.rs
matbd: Not sure if this comment shows approval or disapproval :)
To clarify, just in case, `vec![Format::unknown(); len]` would be sharing the same variable `len` times instead of creating `len` distinct variables.

serde-reflection/src/format.rs
matbd: So TBH it's public mostly for unit tests.
On second thought, I made the type more precise (although possibly less intuitive) because I found no reason no too.

serde-reflection/src/format.rs
matbd: I tried and unfortunately (or fortunately for type safety of Rust :)), the following code doesn't work: 
```
fn borrow(&self) -> &Option<T> {  // <- & instead of Ref is required by the trait
    self.0.as_ref().borrow().deref() // but now we cannot delay the "unborrowing" of the refcell so lifetime checks fail.
}
```

serde-reflection/src/de.rs
bmwill: Yeah it depends on how Clone is implemented, as it will call clone on the type provided `len` times.

serde-reflection/src/format.rs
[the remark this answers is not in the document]
matbd: That worked.

serde-reflection/src/format.rs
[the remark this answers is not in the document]
matbd: done

serde-reflection/src/de.rs
matbd: Right. Here, we want `Clone` to create sharing.

PR #11, opened 2020-05-26
Use String rather than &'static str in Registry

Ploppz:
What I'm trying to do with serde-reflection (providing a proc-macro alternative to generate `ContainerFormat`) does not work if the key of BTreeMap has to be `'static`.

### DIFF
--- a/serde-reflection/src/trace.rs
+++ b/serde-reflection/src/trace.rs
@@ -12,10 +12,7 @@ use serde::{de::DeserializeSeed, Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 
 /// A map of container formats.
-pub type Registry = BTreeMap<&'static str, ContainerFormat>;
-
-/// Similar to `Registry` but compatible with `serde::de::DeserializeOwned`.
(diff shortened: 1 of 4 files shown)


serde-reflection/src/trace.rs
```diff
@@ -12,7 +12,7 @@ use serde::{de::DeserializeSeed, Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 
 /// A map of container formats.
-pub type Registry = BTreeMap<&'static str, ContainerFormat>;
+pub type Registry = BTreeMap<String, ContainerFormat>;
 
 /// Similar to `Registry` but compatible with `serde::de::DeserializeOwned`.
 pub type RegistryOwned = BTreeMap<String, ContainerFormat>;
```
matbd, line 18: Can you remove this one?

Ploppz: Ah, didn't see it. Done.

matbd:
Thanks. I'm ok with the change as it simplifies things a little for everyone.

PR #13, opened 2020-05-28
[Serde generate] python code generation

matbd:
## Summary

This PR creates a new tool `serde-generate` providing the following initial features for python:
* Generate python classes based on Serde formats
* Communicate with Rust in Bincode using an experimental python runtime. (All Serde formats are supported except `char`, `f32`, `f64`.)
* Utility crates `test-utils` provides formats and sample values to ensure full test coverage of Serde constructions. (This is meant to be re-used for other languages and possibly other encodings in the future.)

The python runtime for bincode was based on the initial work of @thefallentree for the LCS encoding. Both the runtime and the generated python code are considered experimental and likely to evolve in the future.

## Test Plan

```
cargo test
cargo run -p serde-generate -- <(curl https://raw.githubusercontent.com/libra/libra/master/testsuite/generate-format/tests/staged/libra.yaml)
```


thefallentree:
could you run python things through `black` ?

serde-generate/runtime/python/bincode/__init__.py
```diff
@@ -0,0 +1,227 @@
...
+    st.int16: lambda content: (st.int16(int.from_bytes(content[:2], byteorder="little", signed=True)), content[2:]),
+    st.int32: lambda content: (st.int32(int.from_bytes(content[:4], byteorder="little", signed=True)), content[4:]),
+    st.int64: lambda content: (st.int64(int.from_bytes(content[:8], byteorder="little", signed=True)), content[8:]),
+    st.int128: lambda content: (st.int128(int.from_bytes(content[:16], byteorder="little", signed=True)), content[16:]),
+
+    st.float32: lambda content: not_implemented(),
+    st.float64: lambda content: not_implemented(),
+
+    st.unit: lambda content: (None, content),
+    st.char: lambda content: not_implemented(),
```
thefallentree, line 95: this could be just chr(content[0]), content[1:]

matbd: No, unfortunately, `char` is supposed to be a unicode character -- taking between 1 and 4 bytes depending on the value.

PR #14, opened 2020-05-29
[serde generate] add support for rust

matbd:
## Summary

* Generate Rust code from the Serde format. (One possible use case is for documentation)
* End-to-end to verify interoperability using bincode encoding.

## Test Plan

```
cargo test
cargo run -p serde-generate -- --language rust <(curl https://raw.githubusercontent.com/libra/libra/master/testsuite/generate-format/tests/staged/libra.yaml)
```

### DIFF
--- a/serde-generate/src/rust.rs
+++ b/serde-generate/src/rust.rs
@@ -1,0 +1,204 @@
+// Copyright (c) Facebook, Inc. and its affiliates
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::analyzer;
+use serde_reflection::{ContainerFormat, Format, Named, Registry, VariantFormat};
+use std::collections::{BTreeMap, HashSet};
+use std::io::{Result, Write};
+
+pub fn output(
+    out: &mut dyn Write,
+    with_derive_macros: bool,
+    registry: &Registry,
+) -> std::result::Result<(), Box<dyn std::error::Error>> {
+    let dependencies = analyzer::get_dependency_map(registry)?;
+    let entries = analyzer::best_effort_topological_sort(&dependencies);
+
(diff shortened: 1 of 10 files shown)


serde-generate/src/rust.rs
```diff
@@ -0,0 +1,204 @@
...
+        F32 => "f32".into(),
+        F64 => "f64".into(),
+        Char => "char".into(),
+        Str => "String".into(),
+        Bytes => "ByteBuf".into(),
+
+        Option(format) => format!("Option<{}>", quote_type(format, known_sizes)),
+        Seq(format) => format!("Vec<{}>", quote_type(format, None)),
+        Map { key, value } => format!(
+            "BTreeMap<{}, {}>",
```
bmwill, line 78: Does this have to be  BTreeMap?

serde-generate/src/rust.rs
```diff
@@ -0,0 +1,204 @@
...
+use serde_bytes::ByteBuf;"#
+    } else {
+        // for testing
+        "type ByteBuf = Vec<u8>;"
+    };
+    writeln!(
+        out,
+        r#"
+#![allow(unused_imports)]{}
+use std::collections::BTreeMap;
```
bmwill, line 40: Maybe you could use the type directly with a fully qualified path instead of having to allow unused imports?

matbd: Good idea. Let me simplify this.

serde-generate/src/rust.rs
matbd: Well, having a stable ordering is definitely more convenient for testing with bincode. I suppose we can always add an option later if needed.

PR #16, opened 2020-06-11
[serde generate] Add options to install module and runtime(s)

matbd:
## Summary

* Use `include_str!` to embed serde types and the bincode runtime of python into `serdegen`.
* Add options to control how to install a rust crate or python modules for the generated code and the embedded runtimes.

## Test Plan

new unit tests

```
cargo run -p serde-generate -- --with-runtimes serde bincode --name libra_types --target-source-dir ../testing --language python3 <(curl https://raw.githubusercontent.com/libra/libra/master/testsuite/generate-format/tests/staged/libra.yaml) &&
(cd ../testing && python3 -c 'import libra_types as lt; import bincode; help(lt.Transaction)')
```

### DIFF
--- a/serde-generate/src/generate.rs
+++ b/serde-generate/src/generate.rs
@@ -7,7 +7,7 @@
 //! cargo run -p serde-generate -- --help
 //! '''
 
-use serde_generate::{python3, rust};
+use serde_generate::{python3, rust, SourceInstaller};
 use serde_reflection::Registry;
 use std::path::PathBuf;
 use structopt::{clap::arg_enum, StructOpt};
@@ -20,6 +20,14 @@ enum Language {
 }
 }
 
+arg_enum! {
+#[derive(Debug, StructOpt, PartialEq, Eq, PartialOrd, Ord)]
+enum Runtime {
+    Serde,
+    Bincode,
+}
+}
+
 #[derive(Debug, StructOpt)]
 #[structopt(
     name = "Serde code generator",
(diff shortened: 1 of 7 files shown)


serde-generate/src/generate.rs
```diff
@@ -20,29 +20,88 @@ enum Language {
...
+
 #[derive(Debug, StructOpt)]
 #[structopt(
     name = "Serde code generator",
     about = "Generate code for Serde containers"
 )]
 struct Options {
     #[structopt(parse(from_os_str))]
-    input: PathBuf,
+    input: Option<PathBuf>,
```
bmwill, line 38: Would this not be a breaking api change?

serde-generate/src/generate.rs
```diff
@@ -20,29 +20,88 @@ enum Language {
...
-    let registry = serde_yaml::from_str::<Registry>(content.as_str()).unwrap();
+    let named_registry_opt = match &options.input {
+        None => None,
+        Some(input) => {
+            let name = options.name.clone().unwrap_or_else(|| {
+                input
+                    .file_stem()
+                    .expect("failed to deduce module name from input path")
+                    .to_string_lossy()
+                    .to_string()
```
bmwill, line 63: I don't believe you need both of these calls? ...Oh it returns a Cow, might be clearer to write `into_owned` then.

serde-generate/src/lib.rs
```diff
@@ -109,3 +109,19 @@ pub mod rust;
 #[doc(hidden)]
 /// Utility functions to help testing code generators.
 pub mod test_utils;
+
+/// How to copy generated source code and available runtimes for a given language.
+pub trait SourceInstaller {
+    /// Create a module exposing the container types contained in the registry.
```
matbd, line 115: On second thoughts, I should probably add a `type Error` in the trait.

serde-generate/src/generate.rs
```diff
@@ -20,29 +20,88 @@ enum Language {
...
+            });
+            let source_path = input.as_os_str();
```
bmwill, line 65: This shouldn't be needed as `read_to_string` takes anything that impls AsRef<Path>

serde-generate/src/generate.rs
```diff
@@ -20,29 +20,88 @@ enum Language {
...
+                std::fs::read_to_string(source_path).expect("input file must be readable");
+            let registry = serde_yaml::from_str::<Registry>(content.as_str()).unwrap();
+            Some((registry, name))
+        }
+    };
+
+    match options.target_source_dir {
+        None => {
+            if let Some((registry, _)) = named_registry_opt {
+                let mut out = std::io::stdout();
```
bmwill, line 76: For performance you may want to lock stdout (assuming you don't need to write out to stdout anywhere else)

serde-generate/src/python3.rs
```diff
@@ -184,3 +185,49 @@ fn output_container(out: &mut dyn Write, name: &str, format: &ContainerFormat) -
...
+}
+
+impl Installer {
+    pub fn new(install_dir: PathBuf) -> Self {
+        Installer { install_dir }
+    }
+
+    fn open_module_init_file(&self, name: &str) -> Result<std::fs::File> {
+        let dir_path = self.install_dir.join(name);
+        std::fs::create_dir_all(dir_path.clone())?;
```
bmwill, line 200:
```suggestion
        std::fs::create_dir_all(&dir_path)?;
```

serde-generate/src/rust.rs
```diff
@@ -190,3 +191,57 @@ fn output_container(
...
+    install_dir: PathBuf,
+}
+
+impl Installer {
+    pub fn new(install_dir: PathBuf) -> Self {
+        Installer { install_dir }
+    }
+
+    fn runtimes_not_implemented() -> Result<()> {
+        Err(std::io::Error::new(
```
bmwill, line 205: It doesn't change too much but you can always bring into scope `use std::io` or `use std::fs` so that you can drop the `std` everywhere. that's mostly a personal preference though

serde-generate/src/rust.rs
```diff
@@ -190,3 +191,57 @@ fn output_container(
...
+}
+
+impl crate::SourceInstaller for Installer {
+    fn install_module(
+        &self,
+        name: &str,
+        registry: &Registry,
+    ) -> std::result::Result<(), Box<dyn std::error::Error>> {
+        let dir_path = self.install_dir.join(name);
+        std::fs::create_dir_all(dir_path.clone())?;
```
bmwill, line 219:
```suggestion
        std::fs::create_dir_all(&dir_path)?;
```

serde-generate/src/generate.rs
matbd: The CLI argument is now optional, so I'm "adding" a feature, no?

serde-generate/src/rust.rs
matbd: Yeah, well, I usually try to stick with `std::io::Error` or `Error` but not `io::Error`. Also FWIW I used to `std::` everywhere in C++.

serde-generate/src/generate.rs
matbd: It worked! nice

serde-generate/src/generate.rs
matbd: Anyways, bumping to `0.2.0` looks a bit cleaner.

PR #17, opened 2020-06-16
[serde generate] LCS runtime

matbd:
## Summary

Make available a python LCS runtime as part of Serde generate.

## Test Plan

Included unit tests.

### DIFF
--- a/serde-generate/Cargo.toml
+++ b/serde-generate/Cargo.toml
@@ -17,7 +17,7 @@ exclude = [
 
 [dependencies]
 maplit = "1.0.2"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.112", features = ["derive"] }
 serde_bytes = "0.11.3"
 serde_yaml = "0.8"
 structopt = "0.3.12"
@@ -28,7 +28,7 @@ serde-reflection = { path = "../serde-reflection", version = "0.3.0" }
 tempfile = "3.1"
 bincode = "1.2"
 hex = "0.4.2"
-
+libra-canonical-serialization = { git = "https://github.com/libra/libra.git", branch = "testnet" }
 
 [[bin]]
 name = "serdegen"
--- a/serde-generate/Cargo.toml
+++ b/serde-generate/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.2.0"
+version = "0.2.1"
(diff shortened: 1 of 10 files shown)


serde-generate/Cargo.toml
```diff
@@ -28,7 +28,7 @@ serde-reflection = { path = "../serde-reflection", version = "0.3.0" }
 tempfile = "3.1"
 bincode = "1.2"
 hex = "0.4.2"
-
+libra-canonical-serialization = { git = "https://github.com/libra/libra.git", branch = "master" }
```
thefallentree, line 31: don't we need testnet branch? otherwise one breaking change (LCS related)  will cause trouble for us.

thefallentree:
my suggestion is that we can make serde-generate 
1.  generate an bunch of fixtures files (random LCS struct with LCS bytes)  
2.  install the generated python lcs runtime
3.  verify the python lcs runtime can deserialize and serialize back to exact bytes

it seems to be an simpler and more generic solution for any runtime we want to support

serde-generate/Cargo.toml
matbd: Sure, I can change that.

serde-generate/tests/lcs_runtime.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[test]
+fn test_python_lcs_runtime_on_all_supported_types() {
+    let registry = test_utils::get_registry().unwrap();
+    let dir = tempdir().unwrap();
+    let source_path = dir.path().join("test.py");
+    let mut source = File::create(&source_path).unwrap();
+    python3::output(&mut source, &registry).unwrap();
+
+    let values = test_utils::get_sample_values();
```
matbd, line 94: @thefallentree What you are suggesting is happening here and below. The values are not random at the moment but they cover every LCS construction.

matbd: We test the installer separately (verifying that moving files around does not break python import).

serde-generate/Cargo.toml
matbd: done

serde-generate/tests/lcs_runtime.rs
matbd: This being said, we could do better on testing. In particular, it's missing negative examples.

PR #31, opened 2020-07-30
generate: introduce and use an indenting writer

bmwill:


serde-generate/src/indent.rs
```diff
@@ -0,0 +1,135 @@
...
+        self.dec_level_by(1);
+    }
+
+    pub fn inc_level_by(&mut self, n: usize) {
+        self.indent_level = self.indent_level.saturating_add(n);
+    }
+
+    pub fn dec_level_by(&mut self, n: usize) {
+        self.indent_level = self.indent_level.saturating_sub(n);
+    }
```
matbd, line 48: I think we only need `inc_level` (aka "indent" ?) and `dec_level` ("unindent"?) in practice. Also, it would be really nice to find a trick to remove `dec_level` and use RAII (scoping) instead.

serde-generate/src/java.rs
```diff
@@ -1,11 +1,14 @@
 // Copyright (c) Facebook, Inc. and its affiliates
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use crate::indent::{Indent, IndentedWriter};
 use include_dir::include_dir as include_directory;
 use serde_reflection::{ContainerFormat, Format, FormatHolder, Named, Registry, VariantFormat};
-use std::collections::BTreeMap;
-use std::io::{Result, Write};
-use std::path::PathBuf;
+use std::{
```
matbd, line 7: Is there an option we can hardcode in CI for `cargo fmt`?

serde-generate/src/java.rs
```diff
@@ -485,28 +488,29 @@ fn output_struct_or_variant_container(
     nested_class: bool,
     package_prefix: &str,
 ) -> Result<()> {
-    let tab = " ".repeat(indentation);
+    let mut out = IndentedWriter::new(out, Indent::Space(4));
+    out.set_level(indentation / 4);
```
matbd, line 492: Now or later, we can simplify the code further by replacing all `out: &dyn Write` occurrences by an indented writer.

serde-generate/src/java.rs
```diff
@@ -485,28 +488,29 @@ fn output_struct_or_variant_container(
...
     // Beginning of class
     if let Some(base) = variant_base {
+        writeln!(out)?;
         writeln!(
             out,
-            "\n{}public static final class {} extends {} {{",
```
matbd, line 493: I don't think you can remove the empty new line yet.

serde-generate/src/lib.rs
```diff
@@ -119,6 +119,8 @@ pub mod python3;
 /// Support for code-generation in Rust
 pub mod rust;
 
+mod indent;
```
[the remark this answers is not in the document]
bmwill: I'll leave that to you to make it public. I don't want to interfere in any release process you have

serde-generate/src/indent.rs
```diff
@@ -0,0 +1,135 @@
...
+            let (before_newline, newline, after_newline) =
+                if let Some(idx) = buf.iter().position(|&b| b == b'\n') {
+                    (&buf[..idx], true, &buf[idx + 1..])
+                } else {
+                    (buf, false, &buf[buf.len()..])
+                };
+
+            if self.at_begining_of_line && !before_newline.is_empty() {
+                for _ in 0..self.level() {
+                    match self.indent {
```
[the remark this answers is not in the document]
bmwill: Its probably not a terrible perf hit doing it this way. If we find it is bad we can fix it like how you've described (though it would take a little extra tracking.

serde-generate/src/indent.rs
bmwill: In practice that's true. of course right now I'm using `set_level` since it hasn't been piped through everywhere yet. Perhaps once we get used to using this we play with RAII and removing these extra methods?

serde-generate/src/java.rs
bmwill: I added a new `writeln!` call jus before this, in order to make it clearer

serde-generate/src/indent.rs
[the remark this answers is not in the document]
matbd: It should be an easier call once the APIs `set_level`, `inc_level_by`, `dec_level_by`... are minimized.

serde-generate/src/indent.rs
matbd: We can certainly experiment with one file (say java.rs) but I would rather refactor the rest of the files only once, with the final (hopefully idiot-proof) API.

matbd: @bmwill I tried some refactoring on my side: https://github.com/facebookincubator/serde-reflection/pull/33. RAII is not playing too well with `&mut out`. Also, all our errors are final (uncaught) so maybe we just don't need this.

bmwill: when you say "this" are you refering to RAII?

matbd: Yes. I agree with you that it's simpler to try RAII at the very end.

PR #33, opened 2020-07-31
[serde generate] clean up codegen

matbd:
## Summary

Make code prettier before it gets more complicated.

## Test Plan

unit tests


matbd:
will reopen a new one

PR #39, opened 2020-08-10
[Serde generate] configuration objects for codegen in all languages

matbd:
## Summary

Advance use-cases in Libra (Move ABIs) require adding more and more codegen options. After this PR, we should be able to do so without breaking the public API (or duplicating functions).

I also simplified some of the options in the process:
* requiring an install directory for Java (i.e. un-support the nested-class trick to output a standalone file),
* making the "module name" mandatory although it is not strictly needed for standalone files in python and Rust.

Comments and external definitions are now supported in all languages.

I also stopped the code-duplication madness between LCS and bincode runtime tests.

Remove code-duplication in Python codegen.

## Test Plan

included tests

Closes https://github.com/facebookincubator/serde-reflection/issues/32


serde-generate/src/rust.rs
```diff
@@ -167,7 +187,7 @@ where
             }
         }
         writeln!(self.out)?;
-        if !self.with_serialization && !external_names.contains("Bytes") {
+        if !self.config.inner.serialization && !external_names.contains("Bytes") {
```
bmwill, line 190: Whats with the exclusion with the Bytes type?

serde-generate/src/python3.rs
```diff
@@ -94,38 +149,52 @@ import typing
...
             .join(", ")
     }
 
+    fn output_comment(&mut self, name: &str) -> std::io::Result<()> {
+        let mut path = self.current_namespace.clone();
+        path.push(name.to_string());
+        if let Some(doc) = self.config.inner.comments.get(&path) {
+            write!(
+                self.out,
+                r#"""" {}
```
bmwill, line 183: This might be easier to read with an explicit `\n` but its just preference.

serde-generate/src/cpp.rs
```diff
@@ -111,23 +129,46 @@ where
...
-    ) -> String {
+    fn enter_class(&mut self, name: &str) {
+        self.out.indent();
+        self.current_namespace.push(name.to_string());
+    }
+
+    fn leave_class(&mut self) {
+        self.out.unindent();
+        self.current_namespace.pop();
+    }
```
bmwill, line 140: 😄

serde-generate/src/generate.rs
```diff
@@ -94,7 +94,7 @@ fn main() {
                         rust::output(&mut out, /* with_derive_macros */ true, &registry).unwrap()
                     }
                     Language::Cpp => cpp::output(&mut out, &registry, Some(&name)).unwrap(),
-                    Language::Java => java::output(&mut out, &registry, &name).unwrap(),
+                    Language::Java => panic!("Code generation in Java requires `--install-dir`"),
```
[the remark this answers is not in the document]
matbd: The Java compiler wants one class per source file :)
We can (and did) work around it using nested classes but it's ugly.

serde-generate/src/java.rs
```diff
@@ -11,24 +10,21 @@ use std::{
...
-    serialization: bool,
-    external_definitions: ExternalDefinitions,
-    comments: DocComments,
+/// Main configuration object for code-generation in Java.
+pub struct JavaCodegenConfig<'a> {
+    /// Language-independent configuration.
+    inner: &'a crate::CodegenConfig,
+    /// Mapping from external type names to fully-qualified class names (e.g. "MyClass" -> "com.facebook.my_package.MyClass").
+    /// Derived from `config.external_definitions`.
+    external_qualified_names: HashMap<String, String>,
```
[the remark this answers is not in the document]
matbd: Right so this field is set by the constructor and not public facing.

serde-generate/src/rust.rs
matbd: This is a hack to allow overriding the definition of `Bytes`. There is no universally good choice unfortunately between `Bytes = Vec<u8>` (doesn't call the `bytes` callback in Serde) and `Bytes = serde_buf::ByteBuf` (creates one user-visible level of indirection). By default, we use the first choice when `serialization` is false and the second otherwise.

serde-generate/src/cpp.rs
matbd: oh, pun not indeed :)

serde-generate/src/rust.rs
bmwill: Oh right! I forgot about that bit :/

PR #41, opened 2020-08-13
[Serde generate] improve python codegen

matbd:
## Summary

* Fix formatting of comments
* omit INDEX and VARIANTS when serialization is skipped
* (possibly breaking) set `# pyre-strict` in generated code

## Test Plan

include unit tests + test in Libra

### DIFF
--- a/serde-generate/Cargo.toml
+++ b/serde-generate/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.10.0"
+version = "0.11.0"
 description = "Generate (de)serialization code in multiple languages"
 documentation = "https://docs.rs/serde-generate"
 repository = "https://github.com/facebookincubator/serde-reflection"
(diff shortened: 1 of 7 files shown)


serde-generate/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.10.0"
+version = "0.11.0"
```
bmwill, line 3: This doesn't appear to be a breaking API change? maybe you can just do a point release?

matbd: I wonder if adding `#pyre-strict` could trigger a chain reaction with more typing errors in python user code.

matbd: I guess it depends if `breaking` is defined in terms of Rust compilation and behaviors only :)

bmwill: eh fair point :) I was strictly thinking in terms of the crate's API but i suppose "breaking" has other  implications as well.

Issue #42, opened 2020-08-14
[Feature Request] LCS/bincode runtime for golang


matbd:
# 🚀 Feature Request

## Motivation

This is part of https://github.com/facebookincubator/serde-reflection/issues/37

## Pitch

We are implementing Go support in a similar way to Java. The callbacks look like this in Go:
https://github.com/facebookincubator/serde-reflection/blob/master/serde-generate/runtime/golang/src/serde/interfaces.go

This is a low-level API (and no comments yet, sorry) but you may check the Java code for reference:
https://github.com/facebookincubator/serde-reflection/tree/master/serde-generate/runtime/java/com/facebook/serde
https://github.com/facebookincubator/serde-reflection/tree/master/serde-generate/runtime/java/com/facebook/lcs
https://github.com/facebookincubator/serde-reflection/tree/master/serde-generate/runtime/java/com/facebook/bincode

In Java, we shared most of the code between bincode and LCS using a Binary(De)Serializer abstraction. It would be really nice to do the same here. (This gives a bincode implementation basically for free, without LCS complications with maps.)

In terms of testing, you may write unit tests in golang and run them from a Rust test like this:
https://github.com/facebookincubator/serde-reflection/blob/master/serde-generate/runtime/java/com/facebook/lcs/LcsTest.java
https://github.com/facebookincubator/serde-reflection/blob/master/serde-generate/tests/runtimes.rs#L679

There will also be one or several end-to-end tests (when I'm done with codegen) driven by a Rust test, like this:
https://github.com/facebookincubator/serde-reflection/blob/master/serde-generate/tests/runtimes.rs#L568

PR #46, opened 2020-08-15
golang lcs serializer & deserializer implementation round one

xli:
First round implementing golang lcs serializer & deserializer.
Map is not supported yet, will have follow up change to implement.

Note: added go module under golang/src for now, otherwise we need use "github.com/facebookincubator/serde-reflection/src/lcs" instead of "github.com/facebookincubator/serde-reflection/lcs" to import, maybe we should remove "src", seems not necessary in golang.

#42

Test Plan:

unit test


matbd:
Currently, I use GOPATH to compile from rust. I read that `src` was the way to go: https://golang.org/pkg/go/build/

It looks like you need to tweak the codegen that uses GetBufferOffset.

serde-generate/runtime/golang/src/serde/interfaces.go
```diff
@@ -88,11 +88,11 @@ type Deserializer interface {
 
 	DeserializeLen() (int, error)
```
matbd, line 89: My mistake, it should have been `uint64`

serde-generate/runtime/golang/src/serde/interfaces.go
```diff
@@ -88,11 +88,11 @@ type Deserializer interface {
...
 
-	DeserializeVariantIndex() (uint64, error)
+	DeserializeVariantIndex() (uint32, error)
 
 	DeserializeOptionTag() (bool, error)
 
-	GetBufferOffset() uint64
+	GetBufferOffset() int
```
matbd, line 95: Actually, `uint64` is better.

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
...
+	len, err := d.DeserializeLen()
+	if err != nil {
+		return nil, err
+	}
+	ret := make([]byte, len)
+	_, err = d.buf.Read(ret)
+	return ret, err
+}
+
+func (d *Deserializer) DeserializeLen() (int, error) {
```
matbd, line 38: `uint32` makes sense for LCS but in general (bincode) it should be `uint64` (my bad). Java was an exception because it has no signed integers and arrays are limited to 2^31 elements.

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
...
+	return string(bytes), err
+}
+
+func (d *Deserializer) DeserializeBool() (bool, error) {
+	ret, err := d.buf.ReadByte()
+	return ret == 1, err
+}
+
+func (d *Deserializer) DeserializeUnit() (struct{}, error) {
+	return struct{}{}, nil
```
matbd, line 57: Doesn't it work with a single `{}`?

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// DeserializeF64 is unimplemented
+func (d *Deserializer) DeserializeF64() (float64, error) {
+	panic("unimplemented")
+}
+
+func (d *Deserializer) DeserializeU8() (uint8, error) {
+	ret, err := d.buf.ReadByte()
+	return uint8(ret), err
```
matbd, line 77: `return d.buf.ReadByte()` should work

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
...
+	return d.buf.Len()
+}
+
+// CheckThatKeySlicesAreIncreasing is unimplemented yet
+func (d *Deserializer) CheckThatKeySlicesAreIncreasing(key1, key2 serde.Slice) error {
+	panic("unimplemented")
+}
+
+func (d *Deserializer) deserializeUleb128AsU32() (uint32, error) {
+	value := uint64(0)
```
matbd, line 178: perhaps `var value uint64 = 0` then?

serde-generate/runtime/golang/src/lcs/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+func NewSerializer() serde.Serializer {
+	return new(Serializer)
+}
+
+func (s *Serializer) GetBytes() []byte {
+	return s.buf.Bytes()
+}
+
+func (s *Serializer) SerializeLen(value int) error {
+	if value < 0 {
```
matbd, line 28: `value` should be uint64 actually so the check is more `value > MaxLen` => error.

serde-generate/runtime/golang/src/lcs/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+		return errors.New("length must >= 0")
+	}
+	s.serializeU32AsUleb128(uint32(value))
+	return nil
+}
+
+func (s *Serializer) SerializeBytes(value []byte) error {
+	s.SerializeLen(len(value))
+	s.buf.Write(value)
+	return nil
```
matbd, line 38: Sadly, each of those `s.buf.Write(value)` calls can fail in theory. Not sure if it's worth the trouble. (Looking at the docs, it seems like it won't panic in case of OOM)

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
...
+	return ret == 1, err
+}
+
+func (d *Deserializer) DeserializeUnit() (struct{}, error) {
+	return struct{}{}, nil
+}
+
+// DeserializeChar is unimplemented
+func (d *Deserializer) DeserializeChar() (rune, error) {
+	panic("unimplemented")
```
matbd, line 62: Perhaps better throw an error than panicking, since we can.

serde-generate/src/golang.rs
```diff
@@ -443,10 +443,10 @@ obj := make(map[{0}]{1})
 previous_slice := serde.Slice {{ 0, 0 }}
 for i := 0; i < length; i++ {{
 	var slice serde.Slice
-	slice.Start = deserializer.GetBufferOffset()
+	slice.Start = uint64(deserializer.GetBufferOffset())
```
matbd, line 446: It looks like the definition of `serde.Slice` is right and `deserializer.GetBufferOffset()` should be an `uint64`. (Once again Java is stuck in a 32-bit world and gave us the wrong impression.)

serde-generate/runtime/golang/src/lcs/serializer.go
xli: Write appends the contents of p to the buffer, growing the buffer as needed. The return value n is the length of p; err is always nil. If the buffer becomes too large, Write will panic with ErrTooLarge.

Intend to return nil and tell user, you can ignore the error returned, in case you really want to cover all branches (test error case)

serde-generate/runtime/golang/src/lcs/deserializer.go
xli: this is for keep it consistent with SortMapEntries, which does not have error returned.

serde-generate/src/golang.rs
xli: 1. I don't think it really matters to support uint64, as 99.99999999999999999% cases will not hit int limit, because it means 2^31-1 bytes, it's 2GB data, and it probably not going to performance well.
2. as I used bytes.Buffer internally, the Buffer#Len is int type, if we changed to uint64, it basically means we need re-impl bytes.Buffer, given the above 1, I don't think it worth the effort.

serde-generate/runtime/golang/src/lcs/serializer.go
xli: by the way, I prefer test all cases including errors.

serde-generate/runtime/golang/src/lcs/deserializer.go
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Facebook, Inc. and its affiliates
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+package lcs
+
+import (
+	"bytes"
+	"errors"
+
+	"github.com/facebookincubator/serde-reflection/serde"
```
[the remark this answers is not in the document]
xli: It's golang's convention to use a uri to import your package. See https://golang.org/doc/code.html#ImportingRemote .

It has huge benefit, to give you example, I updated my https://github.com/xli/serde-reflection master branch with mod name "github.com/xli/serde-reflection/serde-generate/runtime/golang/src" (see https://github.com/xli/serde-reflection/commit/de760e8ba0674649b2558ce7e91d1b434e87b043)

Then in my libra go codebase import "github.com/xli/serde-reflection/serde-generate/runtime/golang/src/lcs",
and use the lcs.
Without doing anything else, when running test / build, I got:

```bash
go test -run='' .
go: finding module for package github.com/xli/serde-reflection/serde-generate/runtime/golang/src/lcs
go: downloading github.com/xli/serde-reflection v0.0.0-20200815182109-5a0de5262f9e
go: downloading github.com/xli/serde-reflection/serde-generate/runtime/golang/src v0.0.0-20200815182109-5a0de5262f9e
go: found github.com/xli/serde-reflection/serde-generate/runtime/golang/src/lcs in github.com/xli/serde-reflection/serde-generate/runtime/golang/src v0.0.0-20200815182109-5a0de5262f9e
go: downloading github.com/xli/serde-reflection v0.0.0-20200815025256-253d68995d66
ok  	github.com/libra/libra-client-sdk-go/libraclient	(cached)
```

It means, we don't need install this runtime when generating code, we just need import correct path "github.com/facebookincubator/serde-reflection/serde-generate/runtime/golang/src/serde" in generated code, it will just work.
The test golang_generation change proves this.

I have not remove the "src" yet, but if we all agree this is better way to install the runtime, we should remove the "src" under golang.

----------------

Install as serde is not worth it, because it is customized GOPATH, user will run into various problems, and they can't checkin the code into their go repository, because in general, people follow the golang guide to create repository under xxx.com/repo/....
Hence if copy to the repo, it means the serde should be imported as "xxx.com/repo/serde", and it won't work for our own code import "serde".
This will become nightmare for user, as even the data structures are exactly same, but golang consider they are different types, because the import path is different.

serde-generate/runtime/golang/src/lcs/serializer.go
xli: Not worth for uint64 because:
1. when input value is int, it's max value == MaxSequenceLength = (1 << 31) - 1, hence no need to validate overflow, although it requires check value < 0 case, which is same effort with checking > MaxLen
2. it tells wrong information to user about how big of the Len could be. Max int is your limit. Where negative is much easier to understand and less likely will do it as len value comparing with pass in more than MaxSequenceLength value, as caller usually just call len(xxx) to get length in golang.
3. in almost 99.99999999999% cases, it is an array or slice, we do len(xxx) to get length, which is int. Using uint64 means an additional type convert to the caller's code every time calls to this function.

serde-generate/runtime/golang/src/lcs/serializer.go
xli: The following code probably tells everything:

``` golang
func (b *Buffer) Write(p []byte) (n int, err error) {
	b.lastRead = opInvalid
	m, ok := b.tryGrowByReslice(len(p))
	if !ok {
		m = b.grow(len(p))
	}
	return copy(b.buf[m:], p), nil
}
```

xli:
I have to upgrade build golang version to 1.13, because 1.11 has a bug to fetch go mod by replace path in test.
The replace path setup is for local test to find local repo instead of fetching by go mod uri from remote, so this does not mean code / generated code can't be compatible with go 1.11.
Go 1.13 is released one year ago, so it should be good enough for testing.

serde-generate/runtime/golang/src/lcs/deserializer.go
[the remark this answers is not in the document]
matbd: Thanks for the explanations. For users (I mean external repos) indeed it is great if we can use a native package manager and avoid "source install". Brandon, Yucong, and I talked about packaging the python language runtime at some point and deprecate "source install". (It looks like Cmake or Maven also have a notion of external project)

What I don't understand is: when we develop in serde-reflection, how do we make sure that "lcs" refer to the module "serde" of the same branch (PR or local) of the same git repo (github official, forked, or local)?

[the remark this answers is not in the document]
xli: it will refer to same version because they are both under github.com/facebookincubator/serde-reflection/serde-generate/runtime/golang/src, notice the actual download is "github.com/facebookincubator/serde-reflection", and the other parts are under the repo, hence you always get same version, although my go.mod is inside "github.com/facebookincubator/serde-reflection/serde-generate/runtime/golang/src", that's just for local development to recognize the path when my actual repo is at ~/code/serde-reflection/, when go run test / build, it find go.mod, and consider the module name is used instead of path.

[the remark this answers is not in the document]
matbd: Ok so it looks the information we are looking for is in https://golang.org/doc/gopath_code.html (and chapter 10 of the Go book). My understanding is that one needs both (1) suitable urls in package names (for users) and (2) setting GOPATH for local development.

serde-generate/runtime/golang/src/lcs/serializer.go
matbd: indeed

matbd:
> I have to upgrade build golang version to 1.13, because 1.11 has a bug to fetch go mod by replace path in test.
> The replace path setup is for local test to find local repo instead of fetching by go mod uri from remote, so this does not mean code / generated code can't be compatible with go 1.11.
> Go 1.13 is released one year ago, so it should be good enough for testing.

Since we won't download our own packages (see previous comment), I don't think we need this actually.

serde-generate/runtime/golang/src/lcs/deserializer.go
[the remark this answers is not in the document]
xli: Yes for (1), not sure what you mean (2), GOPATH should be set by user as they need. 
For our development, we follow go modules https://blog.golang.org/using-go-modules, go modules are default after go 1.13 release, and it is addition to GOPATH for local development, not necessary required for production as all dependencies can still be pull in by go get into GOPATH, and behave same.

Looks like we have agreement to not use "serde" as import path, I removed "src" under golang dir.

xli:
> > I have to upgrade build golang version to 1.13, because 1.11 has a bug to fetch go mod by replace path in test.
> > The replace path setup is for local test to find local repo instead of fetching by go mod uri from remote, so this does not mean code / generated code can't be compatible with go 1.11.
> > Go 1.13 is released one year ago, so it should be good enough for testing.
> 
> Since we won't download our own packages (see previous comment), I don't think we need this actually.

Not follow here, you mean setup GOPATH for local test? It requires a directory with:

            src/facebookincubator/serde-reflection/serde-generate/runtime/golang

serde-generate/runtime/golang/src/lcs/deserializer.go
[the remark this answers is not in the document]
matbd: I think we are converging but let me schedule a meeting tomorrow to iron out the details

serde-generate/src/golang.rs
```diff
@@ -345,11 +344,11 @@ for _, item := range(value) {{
                 write!(
                     self.out,
                     r#"
-if err := serializer.SerializeLen(len(value)); err != nil {{ return err }}
+if err := serializer.SerializeLen(uint64(len(value))); err != nil {{ return err }}
 offsets := make([]int, len(value))
 count := 0
```
matbd, line 349: Can you make `offsets` an array of uint64 instead of casting below?

serde-generate/src/golang.rs
```diff
@@ -749,26 +734,14 @@ impl crate::SourceInstaller for Installer {
...
     }
 
     fn install_bincode_runtime(&self) -> std::result::Result<(), Self::Error> {
-        /*
-                self.install_runtime(
-                    include_directory!("runtime/golang/src/bincode"),
-                    "src/bincode",
-                )
-        */
         Ok(())
```
matbd, line 741: FYI for Rust, I returned an error saying "use cargo instead"

serde-generate/tests/golang_generation.rs
```diff
@@ -21,18 +21,38 @@ fn test_that_golang_code_compiles_with_config(
...
+        .unwrap();
+    assert!(status.success());
+
+    let mut runtime_mod_path = std::env::current_exe().unwrap();
+    runtime_mod_path = runtime_mod_path.to_owned();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.push("serde-generate/runtime/golang");
```
matbd, line 39: Your formula is a little more precise but I use the following simpler formula everywhere so you may do the same: `std::env::current_dir()
            .unwrap()
            .join("runtime/golang")
            .to_str()
            .unwrap()`

Alternatively, we could create a proper library function.

serde-generate/tests/golang_generation.rs
```diff
@@ -21,18 +21,38 @@ fn test_that_golang_code_compiles_with_config(
...
+
+    let status = Command::new("go")
+        .current_dir(dir.path())
+        .arg("mod")
+        .arg("edit")
+        .arg("-replace")
```
matbd, line 45: nice

serde-generate/src/golang.rs
xli: SortMapEntries(offsets []int), so offsets should be []int, unless you want to change this interface too

PR #47, opened 2020-08-17
add golang runtime tests into build; fix lcs serializer

xli:
## Summary

Broken change added in my previous diff at last min, because runtime golang tests was not run with build.

This diff adds the test and fix the lcs serializer.

## Test Plan

cargo test -p serde-generate --test golang_generation

### DIFF
--- a/serde-generate/runtime/golang/lcs/serializer.go
+++ b/serde-generate/runtime/golang/lcs/serializer.go
@@ -145,7 +145,7 @@ func (s *Serializer) GetBufferOffset() uint64 {
 }
 
 // SortMapEntries is unimplemented.
-func (s *Serializer) SortMapEntries(offsets []int) {
+func (s *Serializer) SortMapEntries(offsets []uint64) {
 	panic("unimplemented")
 }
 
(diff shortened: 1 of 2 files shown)


serde-generate/tests/golang_generation.rs
```diff
@@ -125,3 +125,22 @@ fn test_golang_code_with_external_definitions() {
     let content = std::fs::read_to_string(source_path).unwrap();
     assert!(content.contains("foo.Tree"));
 }
+
+#[test]
+fn test_run_golang_runtime_tests() {
```
matbd, line 130: This goes into `runtime.rs` though

serde-generate/tests/golang_generation.rs
```diff
@@ -125,3 +125,22 @@ fn test_golang_code_with_external_definitions() {
...
+    let mut runtime_mod_path = std::env::current_exe().unwrap();
+    runtime_mod_path = runtime_mod_path.to_owned();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.pop();
+    runtime_mod_path.push("serde-generate/runtime/golang");
```
matbd, line 137: I suspect `.join("../../../../etc")` works too

xli: not work. not sure what's going wrong :)

PR #52, opened 2020-08-25
[serde generate] Go: LCS maps + fixes

matbd:
## Summary

* Support LCS maps
* This makes it possible to use the same testing suite as other languages (covering all LCS types). 
* Fix deserialization of option types
* Add missing LcsSerialize on enums

## Test Plan

unit tests

### DIFF
--- a/serde-generate/runtime/golang/lcs/deserializer.go
+++ b/serde-generate/runtime/golang/lcs/deserializer.go
@@ -18,16 +18,14 @@ const maxUint32 = uint64(^uint32(0))
 
 // Deserializer implements `serde.Deserializer` interface for deserializing LCS serialized bytes.
 type Deserializer struct {
-	initial_size uint64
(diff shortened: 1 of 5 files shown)


serde-generate/runtime/golang/lcs/deserializer.go
```diff
@@ -181,12 +179,14 @@ func (d *Deserializer) DeserializeOptionTag() (bool, error) {
 }
 
 func (d *Deserializer) GetBufferOffset() uint64 {
-	return d.initial_size - uint64(d.buf.Len())
+	return uint64(len(d.input)) - uint64(d.buf.Len())
```
xli, line 182: the input length grows?

matbd: We need to keep the input byte vector `d.input` for checking key map order efficiently.

PR #54, opened 2020-08-26
[serde-generate] Go: bincode runtime

matbd:
## Summary

* Provide a bincode runtime in Go.
* The code is largely shared with LCS, following other language implementations.
* This tentatively concludes the work on adding Go to serde-generate.

## Test Plan

unit test

### DIFF
--- a/serde-generate/runtime/golang/bincode/serializer.go
+++ b/serde-generate/runtime/golang/bincode/serializer.go
@@ -1,0 +1,37 @@
+// Copyright (c) Facebook, Inc. and its affiliates
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+package bincode
+
+import (
+	"github.com/facebookincubator/serde-reflection/serde-generate/runtime/golang/serde"
+)
+
+// `serializer` extends `serde.BinarySerializer` to implement `serde.Serializer`.
+type serializer struct {
+	serde.BinarySerializer
+}
+
+func NewSerializer() serde.Serializer {
+	return new(serializer)
+}
+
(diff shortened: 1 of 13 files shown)


serde-generate/runtime/golang/bincode/serializer.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// `serializer` implements `serde.BinarySerializerInternal`.
+type serializer struct {
+	buf bytes.Buffer
+}
+
+// NewSerializer composes `serializer` and `serde.BinarySerializer` to create a `serde.Serializer`.
+func NewSerializer() serde.Serializer {
+	return serde.NewBinarySerializer(new(serializer))
```
xli, line 19: It's confusing with a BinarySerializer + bincode.Serializer / lcs.Serializer. Reverse order will be better: pass BinarySerializer into bincode.Serializer / lcs.Serializer.
However, I feel BinarySerializer is created purely for sharing some logic, hence it is redundant concept, and I always think bincode == binary code,  hence think they are same thing.

I suggest the following structure:

    bincode.Serializer implements everything in BinarySerializer + what implemented in serializer.
    lcs.Serializer new with a bincode.Serializer instance, and delegate methods it can delegate, implements methods have different implementation.

This way, it is easier to understand what is part of lcs serializer is same with bincode serializer, what part is different.

matbd: I used an embedded struct to simplify the code while sharing a similar amount of logic between LCS and Bincode.

PR #59, opened 2020-09-09
typescript code generator

danprinz:
## Summary

A very simple and naive implementation in TypeScript. 
Should be improved to support (MapEntries, Char, Float32, Float64) in the future.

## Test Plan

Make sure npm is installed on your local system - it should be enough to run the tests


matbd (commented):
@danprinz Thanks for working on this! I just sent you a first batch of comments. Feel free to message me on slack if you have questions.

serde-generate/runtime/ts/serde/binaryDeserializer.ts
```diff
@@ -0,0 +1,113 @@
...
+    r.push(Buffer.from(data));
+    r.push(null);
+    return r;
+  }
+
+  protected constructor(data: Uint8Array) {
+    this.data = BinaryDeserializer.makeReadable(data);
+  }
+
+  abstract deserializeToHexString(): string;
```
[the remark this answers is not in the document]
danprinz: removed

serde-generate/src/generate.rs
```diff
@@ -129,6 +130,9 @@ fn main() {
                         .output(&mut out, &registry)
                         .unwrap(),
                     Language::Java => panic!("Code generation in Java requires `--install-dir`"),
+                    Language::TypeScript => {
+                        panic!("Code generation in TypeScript requires `--install-dir`")
```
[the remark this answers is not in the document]
danprinz: Move everything to a single file, so this constraint is not relevant anymore...

serde-generate/src/lib.rs
```diff
@@ -118,6 +118,8 @@ pub mod java;
 pub mod python3;
 /// Support for code-generation in Rust
 pub mod rust;
+/// Support for code-generation in TypeScript/JavaScript
+pub mod ts;
```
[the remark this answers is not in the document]
danprinz: done

serde-generate/src/ts.rs
```diff
@@ -0,0 +1,1079 @@
...
+use crate::{
+    indent::{IndentConfig, IndentedWriter},
+    CodeGeneratorConfig,
+};
+
+/// Main configuration object for code-generation in TypeScript.
+pub struct CodeGenerator<'a> {
+    /// Language-independent configuration.
+    config: &'a CodeGeneratorConfig,
+    /// Mapping from external type names to fully-qualified class names (e.g. "MyClass" -> "com.facebook.my_package.MyClass").
```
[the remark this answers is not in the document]
danprinz: done

serde-generate/src/ts.rs
```diff
@@ -0,0 +1,1079 @@
...
+            config,
+            external_qualified_names,
+            import_trait_helpers,
+        }
+    }
+
+    /// Output class definitions for ` registry` in separate source files.
+    /// Source files will be created in a subdirectory of `install_dir` corresponding to the given
+    /// package name (if any, otherwise `install_dir` it self).
+    pub fn write_source_files(
```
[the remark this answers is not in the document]
danprinz: only one file now

serde-generate/src/ts.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+    fn output_struct_or_variant_container(
+        &mut self,
+        variant_base: Option<&str>,
+        variant_index: Option<u32>,
+        name: &str,
+        fields: &[Named<Format>],
+    ) -> Result<()> {
+        // add imports to field types
+        if self.generator.import_trait_helpers {
```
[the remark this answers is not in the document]
danprinz: move to Helper static class to be the home for all of the helper functions

serde-generate/src/ts.rs
```diff
@@ -0,0 +1,1079 @@
...
+                )?;
+            }
+
+            _ => panic!("unexpected case"),
+        }
+        self.out.unindent();
+        writeln!(self.out, "}}\n")
+    }
+
+    fn output_typedef_helper(
```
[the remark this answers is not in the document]
danprinz: added

serde-generate/src/ts.rs
```diff
@@ -0,0 +1,1079 @@
...
+        self.enter_class("TraitHelpers", reserved_names);
+        for (mangled_name, subtype) in &subtypes {
+            self.output_serialization_helper(mangled_name, subtype)?;
+            self.output_deserialization_helper(mangled_name, subtype)?;
+        }
+        self.leave_class(reserved_names);
+        writeln!(self.out, "}}\n")
+    }
+
+    fn mangle_type(format: &Format) -> String {
```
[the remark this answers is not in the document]
danprinz: This wouldn't work as the typing system of typescript is completely different. we can try to typedef it for all types, but it would end up pretty much the same, isn't it? or did I miss something?

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,857 @@
...
+                )?;
+            }
+
+            _ => panic!("unexpected case"),
+        }
+        self.out.unindent();
+        writeln!(self.out, "}}\n")
+    }
+
+    /// Make code more readable by using typedefs for complex types (e.g. instead of string | null as a type, we define OptionalString = string | null)
```
matbd, line 519: I'm not sure I understand the rational here. TypeScript supports generic type aliases. Shouldn't we use them instead of code-generating more stuff?
https://www.logicbig.com/tutorials/misc/typescript/generic-type-aliases.html

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,857 @@
...
+            U64 => format!("serializer.serializeU64({}{});", this_str, value),
+            U128 => format!("serializer.serializeU128({}{});", this_str, value),
+            F32 => format!("serializer.serializeF32({}{});", this_str, value),
+            F64 => format!("serializer.serializeF64({}{});", this_str, value),
+            Char => format!("serializer.serializeChar({}{});", this_str, value),
+            Str => format!("serializer.serializeStr({}{});", this_str, value),
+            Bytes => format!("serializer.serializeBytes({}{});", this_str, value),
+            _ => format!(
+                "Helpers.serialize{}({}{}, serializer);",
+                Self::mangle_type(format),
```
matbd, line 280: For the same reason that `serializer.serializeU32` is not the same as `serializer.serializeU16`, we must have that `mangle_type(t1) != mangle_type(t2)` whenever `t1 != t2`. This property is already implemented in `common::mangle_type`. Please use this function and remove `Self::mangle_type`.

serde-generate/runtime/typescript/lcs/lcsDeserializer.ts
```diff
@@ -0,0 +1,31 @@
+import leb from 'leb';
+import { BinaryDeserializer } from '../serde/binaryDeserializer';
+
+export class LcsDeserializer extends BinaryDeserializer {
+  constructor(data: Buffer) {
+    super(data);
+  }
+
+  public deserializeUleb128AsU32(): number {
```
matbd, line 9: This one does not look right.

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,857 @@
...
+            U32 => "number".into(),
+            U64 => "Uint64LE".into(),
+            U128 => "BigNumber".into(),
+            F32 => "number".into(),
+            F64 => "number".into(),
+            Char => "string".into(),
+            Str => "string".into(),
+            Bytes => "Uint8Array".into(),
+
+            Option(format) => format!("Optional{}", self.quote_type(format)),
```
matbd, line 156: Using a generic type alias would be better IMHO. See other comment.

(Incidentally for what you intended to do, it should be `common::mangle_type` here, not `quote_type`.)

matbd (changes requested):
Added a few more comments and explanations.

serde-generate/src/ts.rs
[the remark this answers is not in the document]
matbd: `common::mangle_type` is what you want to generate names indexed by a Serde type. (see other comment)

serde-generate/src/typescript.rs
danprinz: You're right. but the main reason for that is the Tuple type. In Java, it was solved with Tuple2, Tuple3, etc. which I find not so handy. I will change it to Tuple<T extends any[]> to make possible use of Tuple<[string, number]>. it's not so beautiful, but will do the work

matbd: For tuples, we should use typescript native tuples: https://www.tutorialsteacher.com/typescript/typescript-tuple

danprinz: Yes. This is basically what I suggested. I personally think it's a hack (because it actually uses a list), but it would work.

serde-generate/src/typescript.rs
danprinz: There is no function overloading in typescript. this means there is no meaning to the type given to the seralizeX function. for that reason, I think there is no such constraint of mangle_type(t1) != mangle_type(t2) whenever t1 != t2 in regards to serializeX and serializeY.
anyway, I will change it to use only common::mangle_type

serde-generate/runtime/typescript/lcs/lcsDeserializer.ts
danprinz: could you elaborate?

serde-generate/src/typescript.rs
danprinz: will be changed.

serde-generate/src/typescript.rs
matbd: What I meant is that if t1 != t2 then `Helpers.serialize{mangle_type(t1)}` and `Helpers.serialize{mangle_type(t2)}` must be distinct functions because the wire-format are (most likely) different.

serde-generate/runtime/typescript/lcs/lcsDeserializer.ts
matbd: This code doesn't strip the highest bits added by the uleb encoding.

timfish:
Thanks for this PR! 

I spent a couple of hours trying to cobble together a TypeScript solution myself using `syn` and [`ts-binary`](https://www.npmjs.com/package/ts-binary) before I discovered this.

serde-generate/runtime/typescript/lcs/lcsDeserializer.ts
danprinz: the code by itself does not do the actual work. it's up to the "leb" library to do that. 
the thing is that "leb" must get a "clean" buffer with only the needed bytes for decoding. this loop is just for figuring out which bytes to put into that buffer and then let "leb" do the decode.

timfish:
Just been looking through this.

Is there a specific reason why you don't use `BigInt` rather than the `int64-buffer` module?

`BigInt` is built into all modern JavaScript runtimes and plays nicely with `Number` in comparisons and equality so would strike me as more useful to end users.

danprinz:
Well, as I see it there are two options: 
1. Using BigInt - it might result in an unexpected behavior when converting to LE buffers, as by definition "as all BigInts are signed" [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt](see Mozilla docs at the beginning of the operators section)
2. Use BigInt64Array - currently is not supported by Safari [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigUint64Array](under Browser compatibility section)

Do you think I should try using it anyway?
I could do that if not using int64-buffer worth it

timfish:
>  it might result in an unexpected behavior when converting to LE buffers

The conversion from raw bytes should be considered but my concern is that these types are exposed to consumers of the deserialiser and they don't have the ergonomics of `BigInt`. 

Everyone using these serialisers/deserialisers is now required to add `@ethersproject/bignumber`, `int64-buffer` and `@ethersproject/bytes` as dependencies even if they're not used.

`BigInt` can be used with `+`, `-`, `/`, `*`, `>`, `>=`, `<`, `<=` operators and works with comparisons to `number` and as far as I understand, `Uint64LE`, `Int64LE` and `BigNumber` do not support any of these 

So for best ergonomics they would be:
```typescript
deserializeU64(): BigInt;
deserializeU128(): BigInt;
deserializeI64(): BigInt;
deserializeI128(): BigInt;
```
And ideally you'd also want the serialiser to handle `BigInt | number` for these types so that JavaScript objects containing `number` work here. 
```typescript
serializeU64(value: BigInt | number): void;
serializeU128(value: BigInt | number): void;
serializeI64(value: BigInt | number): void;
serializeI128(value: BigInt | number): void;
```

timfish:
There is also mixed usage of `Uint8Array` and `Buffer` and `Buffer` will not work in the browser unless it's shimmed. 

I think much of the Typescript code could be re-written to use only `Uint8Array`, `ArrayBuffer` and [`DataView`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) which are all available in modern browsers and nodejs. You'd just need to manually keep track of the `byteOffset`.

Changing to use `DataView` + `BigInt` means that you could remove the dependencies on `stream`, `@ethersproject/bignumber`, `int64-buffer` and `@ethersproject/bytes`.

timfish:
Sorry for the multiple messages. Here is roughly what I'm suggesting:

```typescript
export abstract class BinaryDeserializer implements Deserializer {
  public view: DataView;
  private index: number;

  protected constructor(data: Uint8Array) {
    this.data = new DataView(data.buffer);
    this.index = 0;
  }

  public deserializeStr(): string {
    const value = this.deserializeBytes();
    return String.fromCharCode.apply(null, Array.from(value));
  }

  public deserializeBool(): boolean {
    const bool = this.view.getUint32(this.index);
    this.index += 4;
    return bool == 1;
  }

  public deserializeUnit(): any {
    return;
  }

  public deserializeU8(): number {
    const value = this.view.getUint8(this.index);
    this.index += 1;
    return value;
  }

  // snip

  public deserializeU64(): BigInt {
    const left = BigInt(dataview.getUint32(this.index);
    const right = BigInt(dataview.getUint32(this.index + 4);
    this.index += 8;

    return (right<<shift?) | left
  }

  // snip

  public deserializeF32(): number {
    const val = this.view.getFloat32(this.index);
    this.index += 4;
    return value;
  }

  public deserializeF64(): number {
    const val = this.view.getFloat64(this.index);
    this.index += 8;
    return value;
  }
}
```

serde-generate/runtime/typescript/serde/binaryDeserializer.ts
```diff
@@ -0,0 +1,113 @@
...
+  abstract deserializeVariantIndex(): number;
+
+  abstract checkThatKeySlicesAreIncreasing(
+      key1: [number, number],
+      key2: [number, number]
+  ): void;
+
+  public deserializeStr(): string {
+    const value = this.deserializeBytes();
+    return String.fromCharCode.apply(null, Array.from(value));
```
timfish, line 34: Text encode/decode should probably go through [`TextEncoder/TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API):
```
var uint8array = new TextEncoder("utf-8").encode("¢");
var string = new TextDecoder("utf-8").decode(uint8array);
```

serde-generate/runtime/typescript/lcs/lcsDeserializer.ts
matbd: Oh I see, my bad

matbd:
> Well, as I see it there are two options:
> 
> 1. Using BigInt - it might result in an unexpected behavior when converting to LE buffers, as by definition "as all BigInts are signed" [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt](see Mozilla docs at the beginning of the operators section)
> 2. Use BigInt64Array - currently is not supported by Safari [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigUint64Array](under Browser compatibility section)
> 
> Do you think I should try using it anyway?
> I could do that if not using int64-buffer worth it

FWIW I used `BigInteger` in Java. One has to be careful, definitely, but you should be able to adapt the Java logics if needed.

matbd (approved):
@danprinz Thanks again for the PR. I have a few more comments and questions but nothing blocking for a first version.

In addition to the elements listed in the summary, to be on par with (say) python or golang, I'd say we are missing:
* unit tests in Typescript,
* runtime tests using the Rust framework (basic (de)serialization first, then including some limited fuzzing),
* the LCS limit on length (`2^31 - 1`) 
* limits on recursion (aka container depth)
* convenient methods `lcsSerialize` and `lcsDeserialize` (and similarly for bincode)

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,747 @@
...
+            U32 => format!("serializer.serializeU32({}{});", this_str, value),
+            U64 => format!("serializer.serializeU64({}{});", this_str, value),
+            U128 => format!("serializer.serializeU128({}{});", this_str, value),
+            F32 => format!("serializer.serializeF32({}{});", this_str, value),
+            F64 => format!("serializer.serializeF64({}{});", this_str, value),
+            Char => format!("serializer.serializeChar({}{});", this_str, value),
+            Str => format!("serializer.serializeStr({}{});", this_str, value),
+            Bytes => format!("serializer.serializeBytes({}{});", this_str, value),
+            _ => format!(
+                "Helpers.serialize{}({}{}, serializer);",
```
matbd, line 234: Just looking at the code, it would be nice to add an underscore here: "Helpers.serialize_{}"

timfish:
This PR was missing the bincode serialiser. There is an empty index.ts where in guessing it should be?

serde-generate/runtime/typescript/serde/binaryDeserializer.ts
```diff
@@ -0,0 +1,129 @@
...
+  public deserializeU16(): number {
+    return new DataView(this.read(2)).getUint16(0, true);
+  }
+
+  public deserializeU32(): number {
+    return new DataView(this.read(4)).getUint32(0, true);
+  }
+
+  public deserializeU64(): BigInt {
+    const left = this.deserializeU32();
```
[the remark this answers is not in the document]
danprinz: I found it very confusing, as when deserializing a little-endian integer, what would be the low bits? the first to be read are the lowest, but when reading it, it's at the high side of the read. I used left and right as an absolute marker. If you think it's better to use high and low, I could change it :-)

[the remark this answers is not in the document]
matbd: I would argue that "low" and "high" are appropriate conventions for a little endian encoding because "low" conveys both the idea of least-significant (low-exponent) bit and the idea of a low address in memory.

Conversely, I'm a bit confused why the least significant part of an integer would be called "left".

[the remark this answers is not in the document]
danprinz: changed to low and high

serde-generate/runtime/typescript/serde/binarySerializer.ts
```diff
@@ -0,0 +1,172 @@
...
+        this.serialize(value);
+    }
+
+    public serializeBool(value: boolean): void {
+        const byteValue = value ? 1 : 0;
+        this.serialize(new Uint8Array([byteValue]));
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars,@typescript-eslint/explicit-module-boundary-types
+    public serializeUnit(value: never): void {
```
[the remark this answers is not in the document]
danprinz: changed

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,747 @@
...
+            external_qualified_names,
+            namespaces_to_import: config
+                .external_definitions
+                .keys()
+                .map(|k| k.to_string())
+                .collect::<Vec<_>>(),
+        }
+    }
+
+    /// Output class definitions for ` registry` in a single source file.
```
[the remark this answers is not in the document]
danprinz: fixed

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,747 @@
...
+                self.out,
+                "import * as {} from '../{}';\n",
+                namespace.to_camel_case(),
+                namespace
+            )?;
+        }
+        writeln!(
+            self.out,
+            r#"
+export type Optional<T> = T | null;
```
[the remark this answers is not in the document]
danprinz: yes. moved it.

serde-generate/runtime/typescript/bincode/bincodeSerializer.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export abstract class BincodeSerializer extends BinarySerializer {
+    serializeLen(value: BigInt): void {
+        this.serializeU64(value);
+    }
+
+    public serializeVariantIndex(value: number): void {
+        this.serializeU32(value);
+    }
+}
```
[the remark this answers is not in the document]
danprinz: done

serde-generate/src/typescript.rs
```diff
@@ -0,0 +1,747 @@
...
+        Ok(())
+    }
+
+    fn quote_type(&self, format: &Format) -> String {
+        use Format::*;
+        match format {
+            TypeName(x) => self.quote_qualified_name(x),
+            Unit => "void".into(),
+            Bool => "boolean".into(),
+            I8 => "number".into(),
```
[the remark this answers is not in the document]
danprinz: done

serde-generate/src/typescript.rs
danprinz: I went with the language most common notation (which is serializeType)

danprinz:
> This PR was missing the bincode serialiser. There is an empty index.ts where in guessing it should be?

I've added Bincode serializer/deserializer

serde-generate/runtime/typescript/serde/binaryDeserializer.ts
```diff
@@ -64,19 +65,19 @@ export abstract class BinaryDeserializer implements Deserializer {
   }
 
   public deserializeU64(): BigInt {
-    const left = this.deserializeU32();
-    const right = this.deserializeU32();
+    const high = this.deserializeU32();
+    const low = this.deserializeU32();
```
matbd, line 69: @danprinz I think you meant the other way around: the first (little-endian-)serialized bytes are called `low`.

danprinz: ok. changed.

PR #61, opened 2020-09-18
[Serde generate] runtimes: negative fuzzing + UTF8 validation + max container depth

matbd:
## Summary

* Use Rust as a baseline to test that supported languages reject incorrect byte strings during LCS deserialization
* Introduce a generated list of "negative" samples for this purpose. Samples are typically obtained by flipping the highest bit of a byte in a valid sample, and testing if Rust can deserialize.
* In all languages, we now also mutate "positive samples" and check that they always decode to a different value (by canonicity)
* The new testing approach covers exceeding input bytes, UTF8 validation, boolean decoding, map key ordering, the length limit of sequences, and the limit on container depth.
* Also, fix a codegen issue in golang for containers of the form `struct Foo(Option<T>)`

## Test Plan

CI

### DIFF
--- a/serde-generate/runtime/golang/serde/binary_deserializer.go
+++ b/serde-generate/runtime/golang/serde/binary_deserializer.go
@@ -7,20 +7,35 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"unicode/utf8"
 )
 
 // `BinaryDeserializer` is a partial implementation of the `Deserializer` interface.
 // It is used as an embedded struct by the Bincode and LCS deserializers.
 type BinaryDeserializer struct {
-	Buffer *bytes.Buffer
-	Input  []byte
+	Buffer               *bytes.Buffer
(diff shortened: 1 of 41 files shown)


serde-generate/runtime/golang/serde/binary_deserializer.go
```diff
@@ -7,36 +7,60 @@ import (
...
 // `deserializeLen` to be provided by the extending struct.
 func (d *BinaryDeserializer) DeserializeStr(deserializeLen func() (uint64, error)) (string, error) {
 	bytes, err := d.DeserializeBytes(deserializeLen)
+	if err != nil {
+		return "", err
+	}
+	if !utf8.Valid(bytes) {
+		return "", errors.New("invalid UTF8 string")
+	}
 	return string(bytes), err
```
xli, line 64: return string(bytes), nil

xli (approved):

PR #62, opened 2020-09-21
[serde generate] fix Java invariants

matbd:
## Summary

* Use proper exception to enforce argument invariants. (Only one "true" assert left.)
* Add missing invariants for `serialize_i128`

## Test Plan

CI

### DIFF
--- a/serde-generate/src/java.rs
+++ b/serde-generate/src/java.rs
@@ -415,9 +415,11 @@ serializer.sort_map_entries(offsets);
                 write!(
                     self.out,
                     r#"
-assert value.length == {};
-for ({} item : value) {{
-    {}
+if (value.length != {0}) {{
+    throw new java.lang.IllegalArgumentException("Invalid length for fixed-size array: " + value.length + " instead of "+ {0});
+}}
+for ({1} item : value) {{
+    {2}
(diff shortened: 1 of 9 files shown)


serde-generate/src/java.rs
```diff
@@ -630,7 +632,11 @@ return obj;
         )?;
         self.out.indent();
         for field in fields {
-            writeln!(self.out, "assert {} != null;", &field.name)?;
+            writeln!(
+                self.out,
+                "java.util.Objects.requireNonNull({});",
```
xli, line 637: ({}, "{} must not be null")

give specific error message with field name

serde-generate/src/java.rs
```diff
@@ -415,7 +415,9 @@ serializer.sort_map_entries(offsets);
                 write!(
                     self.out,
                     r#"
-assert value.length == {};
+if (value.length != {}) {{
+    throw new java.lang.IllegalArgumentException("Invalid length for fixed-size array");
```
xli, line 419: "Invalid length(" + value.length + ") for fixed-size ({})...", add the size to make it specific error message

xli (approved):

PR #63, opened 2020-09-23
[serde generate] java runtime: improve Bytes class

matbd:
## Summary

* Make the default constructor non-copying (does not prevent sharing but faster)
* Create convenient safe APIs `Bytes.empty()` and `Bytes.valueOf(content)`

## Test Plan

CI

### DIFF
--- a/serde-generate/runtime/java/com/novi/serde/Bytes.java
+++ b/serde-generate/runtime/java/com/novi/serde/Bytes.java
@@ -14,9 +14,25 @@ import java.util.Objects;
 public final class Bytes {
     private final byte[] content;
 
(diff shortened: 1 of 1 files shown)


thefallentree:
can we consider use guava instead ? 

https://guava.dev/releases/27.1-jre/api/docs/index.html?com/google/common/primitives/Bytes.html

in general this is the defacto standard for java these days

xli (approved):

matbd:
@thefallentree That's a good question. So far, we've managed to keep the LCS runtimes mostly dependency-free (the only exception I can remember is numpy in python). It seems a bit vexing to pull guava just for Bytes. (FWIW protobuf does not)

PR #66, opened 2020-10-13
[serde generate] add config option `custom_code`

matbd:
## Summary

Add option to inject custom code in various languages.

This is similar to comments, except inserted at the end of a class and without quotes.

## Test Plan

unit test

### DIFF
--- a/serde-generate/src/python3.rs
+++ b/serde-generate/src/python3.rs
@@ -188,6 +188,18 @@ import typing
         Ok(())
     }
 
+    fn output_custom_code(&mut self, name: &str) -> std::io::Result<bool> {
+        let mut path = self.current_namespace.clone();
+        path.push(name.to_string());
+        match self.generator.config.custom_code.get(&path) {
+            Some(code) => {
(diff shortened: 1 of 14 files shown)


serde-generate/src/python3.rs
```diff
@@ -252,7 +268,7 @@ import typing
         writeln!(self.out, "\nclass {}:", name)?;
         self.out.indent();
         self.output_comment(&name)?;
-        if self.generator.config.serialization {
+        if self.generator.config.serialization && !self.has_custom_code(&name) {
```
xli, line 271: why this guard? when has custom code, we lost serialization?

matbd: We output `pass` in the else branch.

xli: or, you think we should just pull out serialization into custom code?

matbd: Oh my bad it should be a OR.

xli (approved):

PR #71, opened 2020-11-13
Implement C# Generation and Runtime

### LINKED ISSUE #68: [Feature Request] Add C# Code Generation
# 🚀 Feature Request

Add code generation support for C#.

## Motivation

C# is a very popular language, which would be nice to support. I need support for it for my own project.

## Pitch

C# is rather similar to Java, so implementing it is a fairly straightforward port of the existing Java code generator.

I'm currently working on implementing this here: https://github.com/mattico/serde-reflection/tree/csharp
I would like to upstream this support if there is a desire for it.

mattico:
## Summary

A fairly straightforward port of the Java code to C#. Targets .NET Standard 1.6 for broad compatibility. Newer .NET versions do have some features that would be useful (`Rune`, `record`, `Range`), maybe we can switch in a few years.

There are a few notable changes compared to the Java version:
1. C-style enums are optionally translated to a C# enum.
2. Wrapper classes are used for maps and arrays to implement value equality.
3. Builder classes are not emitted or used.
4. Code is generated into a single project called `Serde`. Runtimes are in subfolders (and sub-namespaces) of that.

Fixes #68.

## Test Plan

I ported the tests over from Java as well. Let's see if I got CI working on the first try.


matbd:
Thanks for the PR.

Couple of questions/comments while I'm getting started with the code review:
1. I don't think we need to explicitly support hashmaps or hashsets where keys would have different Serde types. Do you have a specific use case in mind?
2. I like this idea a lot but we don't do this in the other languages yet. Can we make this behavior optional (I can add a config option for you) so that we can coordinate a slow rollout across languages? Ideally, this would also be a separate issue and PR in C#.
3. In Java, without a wrapper for Bytes, it's not really feasible to handle composed types like `Map<Bytes, T>`, `(Bytes, T)`, let alone `Map<(Bytes, T), S>`. If the same observation holds for C# (as I would imagine), then it is a strong argument to use a wrapper for byte[]. FWIW protobuf's codegen introduces lots of wrappers around native types in Java and C# (seemingly also to provide a correct "value semantics"): https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/bytes-value
4. Sure, I suppose we can always add builders later. I believe this is the preferred style in Java (and C#?) these days to avoid bugs where arguments of similar types are swapped by mistake.
5. I haven't looked at the details yet but we may have be consistent with java at some point.

mattico:
1. Part of `GetHashCode`s contract is that it is congruent to `Equals`. Given that it is easy to implement correctly I would much rather do that than have to document somewhere that it doesn't work as expected which will inevitably cause problems for someone. I can imagine using an enum as a hashmap key, in my application I was thinking about using a `ConfigKey` enum which would be used as a key into a hashmap. If the concern is the dependency on `rand` I could hash the type or something.
2. Fair enough.
3. That's a good reason to use a wrapper. I had one in an earlier commit which I'll resurrect.
4. I was wondering about the utility of the builder classes. In C# I would use keyword arguments or an object initializer for that.

matbd:
> 1. Part of `GetHashCode`s contract is that it is congruent to `Equals`. Given that it is easy to implement correctly I would much rather do that than have to document somewhere that it doesn't work as expected which will inevitably cause problems for someone. I can imagine using an enum as a hashmap key, in my application I was thinking about using a `ConfigKey` enum which would be used as a key into a hashmap. If the concern is the dependency on `rand` I could hash the type or something.

I believe the only strict requirement is that `x1.equals(x2)` implies `x1.GetHashCode() == x2.GetHashCode()`.

In the examples that I could find over the Net for C# (and Java), people (or IntelliJ) always ignore the class of an object and just consume its fields one by one. (In C#, you have `HashCode.Add` to do so, apparently.)

The proposed improvement to use `rand` at codegen time for the seed is easy and kinda cool but for those who commit the result of the serde-generate in git (like us), it will create meaningless changed lines in git commits.

> 4. I was wondering about the utility of the builder classes. In C# I would use keyword arguments or an object initializer for that.

Sounds good to me!

serde-generate/tests/csharp_runtime.rs
```diff
@@ -0,0 +1,308 @@
...
+        "{{{}}}",
+        bytes
+            .iter()
+            .map(|x| format!("{}", *x as u8))
+            .collect::<Vec<_>>()
+            .join(", ")
+    )
+}
+
+fn test_csharp_runtime_on_supported_types(dir: PathBuf, runtime: Runtime) {
```
matbd, line 213: And now the extended runtime test with fuzzing for both LCS and bincode! Very nice!

matbd (commented):
Thank you for the impressive work! I'll try to make another pass this week. Looking forward to releasing the C# support!

mattico:
> I believe the only strict requirement is that x1.equals(x2) implies x1.GetHashCode() == x2.GetHashCode().
> ...

You're correct, should've validated my assumptions :)

> The proposed improvement to use rand at codegen time for the seed is easy and kinda cool but for those who commit the result of the serde-generate in git (like us), **it will create meaningless changed lines in git commits**. I can live with a solution based on stable hashing of Serde types, but is it worth the effort?

That's a good reason to avoid rand here.  I'll revert that change. We can revisit this if anyone ever needs it.

serde-generate/runtime/csharp/Serde/Option.cs
```diff
@@ -0,0 +1,74 @@
...
+            hashCode = hashCode * -1521134295 + EqualityComparer<T>.Default.GetHashCode(value);
+            return hashCode;
+        }
+
+        public static bool operator ==(Option<T> left, Option<T> right) => Equals(left, right);
+
+        public static bool operator !=(Option<T> left, Option<T> right) => !Equals(left, right);
+    }
+
+    public static class OptionExtensions
```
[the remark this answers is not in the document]
mattico: It's just a style thing. Extension methods can only access the public class members so they're clearly sugar/helper functions, separate from the more fundamental methods of the implementation.

serde-generate/src/csharp.rs
```diff
@@ -0,0 +1,1159 @@
...
+        match format {
+            TypeName(x) => self.quote_qualified_name(x),
+            Unit => "Unit".into(),
+            Bool => "bool".into(),
+            I8 => "sbyte".into(),
+            I16 => "short".into(),
+            I32 => "int".into(),
+            I64 => "long".into(),
+            I128 => "BigInteger".into(),
+            U8 => "byte".into(),
```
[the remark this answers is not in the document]
mattico: I felt a bit bad for Java developers while porting this, spent a lot of time removing boilerplate that was only necessary due to Java's obstinate refusal to add features. Then I remembered all the hours I spent trying to find good open-source C# libraries and stopped being so smug ;)

serde-generate/src/csharp.rs
```diff
@@ -0,0 +1,1159 @@
...
+            Map { key, value } => format!(
+                "Dictionary<{}, {}>",
+                self.quote_type(key),
+                self.quote_type(value)
+            ),
+            Tuple(formats) => format!("({})", self.quote_types(formats)),
+            TupleArray {
+                content,
+                size: _size,
+            } => format!("{}[]", self.quote_type(content),),
```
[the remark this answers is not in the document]
mattico: Could also use `ImmutableArray<T>`.

serde-generate/tests/csharp_runtime.rs
```diff
@@ -0,0 +1,308 @@
...
+use std::sync::Mutex;
+
+lazy_static::lazy_static! {
+    // `dotnet build` spuriously fails on linux if run concurrently
+    static ref MUTEX: Mutex<()> = Mutex::new(());
+}
+
+/// Set to `false` to have tests generate into non-temporary directories for inspection.
+/// NOTE: Set this to `true` before committing or merging!
+const TEST_USE_TEMP_DIRS: bool = true;
```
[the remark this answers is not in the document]
mattico: That's a good alternative.

serde-generate/src/csharp.rs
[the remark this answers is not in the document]
matbd: Right. I meant `java.util.List` in Java.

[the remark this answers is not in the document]
matbd: Ok so I landed the fix in Java: https://github.com/novifinancial/serde-reflection/pull/73

How about porting the new runtime test for self-equality to C#?

mattico:
Rebased, added value-equality test, and added argument parsing for `--c-style-enums`. Not totally happy with the argument parsing, open to suggestions.

serde-generate/tests/csharp_generation.rs
```diff
@@ -0,0 +1,113 @@
...
+#[test]
+fn test_that_csharp_code_compiles_without_serialization() {
+    let config = CodeGeneratorConfig::new("Serde.Generated".to_string()).with_serialization(false);
+    test_that_csharp_code_compiles_with_config(&config);
+}
+
+#[test]
+fn test_that_csharp_code_compiles_with_c_style_enums() {
+    let config = CodeGeneratorConfig::new("Serde.Generated".to_string()).with_c_style_enums(true);
+    test_that_csharp_code_compiles_with_config(&config);
```
matbd, line 58: Not sure there is a test case of C-like enum but I can add it

matbd:
> Rebased, added value-equality test, and added argument parsing for `--c-style-enums`. Not totally happy with the argument parsing, open to suggestions.

Did you meant the `panic!`s? I think that's fine for now. We can always improve later.

I'm going to land this PR now and give us a few days before releasing a crate.

@mattico Thanks for the awesome contribution!

serde-generate/src/generate.rs
```diff
@@ -67,9 +69,13 @@ struct Options {
     /// Optional package name (Python) or module path (Go) where to find Serde runtime dependencies.
     #[structopt(long)]
     serde_package_name: Option<String>,
+
+    /// Translate enums without variant data (c-style enums) into their equivalent in the target language.
+    #[structopt(long)]
+    c_style_enums: bool,
```
matbd, line 75: FYI I may add a verb here later, like `use_c_style_enums`

serde-generate/src/csharp.rs
```diff
@@ -0,0 +1,1170 @@
...
```
matbd, line 283: @mattico Wait, how come this passes the self-equality test?

serde-generate/src/generate.rs
mattico: That does seem more consistent with `--with-runtimes`

serde-generate/tests/csharp_generation.rs
mattico: Done!

serde-generate/src/csharp.rs
mattico: Arrays and other such things use `Enumerable.SequenceEqual`:

https://github.com/novifinancial/serde-reflection/blob/349258eb6b0ddedaeec448cd81e9a4676de0dc61/serde-generate/src/csharp.rs#L827-L840

I just went and looked at the SequenceEqual implementation and it doesn't short-circuit based on the array length, which could be a big deal for `Bytes`. Should use `IStructuralEquatable` instead.

mattico: I'm now realizing that collections which have reference-type elements also need to use the appropriate `IEqualityComparer` for their elements.

mattico:
Also it appears that the test runner stopped being able to discover the tests and it succeeds when there's no tests found :|

serde-generate/src/csharp.rs
matbd: Indeed, `SequenceEqual` does not compose. Also, `GetHashCode` is guaranteed to follow `Equals` (not `SequenceEqual`).

matbd: In Java, `java.util.*` containers have a value-based `equals` and `hashCode` already, so only arrays are problematic (and should avoided) really. Hopefully, it's the same in C#?

mattico: > In Java, java.util.* containers have a value-based equals and hashCode already, so only arrays are problematic (and should avoided) really. Hopefully, it's the same in C#?

Unfortunately, no. No standard container types that I can find implement value-based equality by default. Going to have to generate equality helpers.

matbd: I'm a bit confused. You were already using `ImmutableArray`. How about the other immutable collections for maps etc? https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable

mattico: None of the immutable collections implement value equality either, I was mistaken.

matbd: Ok so using `IStructuralEquatable` and `ImmutableArray`, I managed to make in-depth equality work on 2D matrices: https://gist.github.com/matbd/79e130b5302a93fc3917ecc339809842

It looks like this approach applies to any combination of tuples and arrays (and value types), both for equality and hashing, hence making it possible for those types to be keys in a dictionary: https://stackoverflow.com/questions/3609823/what-problem-does-istructuralequatable-and-istructuralcomparable-solve

This doesn't quite solve the problem of structural equality for user-defined structs `Foo` that would contain a dictionary. For now, I suppose that you could just return a runtime error "Unimplemented" from the implementation of `Equals` and `GetHashCode` of such `Foo`.

Alternatively, at this point, it might be just as easy to define our own collection classes (`Serde.Array`, `Serde.Map`, etc) by deriving (say) the immutable ones and overriding `Equals` and `HashCode` (+ a number of constructors/builders I guess?)

mattico: Yeah it does seem easier to make collection wrappers. It should communicate the behavior better as well.

matbd:
@mattico Thanks the latest changes. I just merged the PR!

matbd:
@mattico fyi I started a PR to prepare the release: https://github.com/novifinancial/serde-reflection/pull/75

PR #73, opened 2020-11-17
[serde-generate] Java: fix equality for tuples and derived types

matbd:
## Summary

* `TupleArray<T, N>` now translates to an `ArrayList<T>` in Java.
* This ensures the value semantics hold for serde values (e.g. for all valid string `s`, we have `deserialize(s) == deserialize(s)`).
* Fix bug in existing runtime test (where I forgot that `==` and `!=` were useless)
* Add test for self-equality.

## Test Plan

CI


serde-generate/src/java.rs
```diff
@@ -221,9 +221,9 @@ where
                 self.quote_types(formats)
             ),
             TupleArray { content, size } => format!(
-                "{} @com.novi.serde.ArrayLen(length={}) []",
-                self.quote_type(content),
-                size
+                "java.util./*@com.novi.serde.ArrayLen(length={})*/List<{}>",
```
xli, line 224: LoL, actually works?!

xli (approved):

serde-generate/src/java.rs
matbd: I got the idea from this thread: https://stackoverflow.com/questions/21385938/java-scoping-construct-cannot-be-annotated-with-type-use

but now I realize we are already using Java >=8 so let me fix that.

PR #76, opened 2020-12-02
[serde-generate] More improvements in C# codegen

matbd:
## Summary

* Consistently use lowercase autogenerated field names "value" and "fieldX" to avoid breaking compilation in case of a collision with a class name (e.g. variant `Value`). This is also more likely to match the rest of the generated code since serde-reflection extracts lowercase field names from Rust. (@mattico We could also add an option to enforce uppercase public fields everywhere, if needed in the future.)
* Change .csproj file again to remove some warnings on my laptop.
* Support generation of C# files outside of the Serde project

## Test Plan

CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,7 +307,7 @@ dependencies = [
 
 [[package]]
 name = "serde-generate"
-version = "0.16.1"
+version = "0.17.0"
 dependencies = [
  "bincode",
  "heck",
--- a/serde-generate/Cargo.toml
+++ b/serde-generate/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.16.1"
+version = "0.17.0"
 description = "Generate (de)serialization code in multiple languages"
 documentation = "https://docs.rs/serde-generate"
 repository = "https://github.com/novifinancial/serde-reflection"
--- a/serde-generate/src/csharp.rs
+++ b/serde-generate/src/csharp.rs
@@ -25,7 +25,7 @@ pub struct CodeGenerator<'a> {
 }
 
 /// Shared state for the code generation of a C# source file.
-struct CsharpEmitter<'a, T> {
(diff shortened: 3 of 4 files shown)


serde-generate/runtime/csharp/Serde.Tests/Serde.Tests.csproj
```diff
@@ -1,21 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
-    <TargetFramework>netcoreapp2</TargetFramework>
-
-    <IsPackable>false</IsPackable>
-
+    <PropertyGroup>
+        <TargetFramework>netstandard1.6</TargetFramework>
```
mattico, line 3: One of my earlier commits used this, but the test runner didn't support loading tests from netstandard DLLs. Can you verify that the tests are actually running?

matbd: Good question.

Sadly, it turns out that on my laptop, nothing is run by `dotnet test` (disregarding this change). Testing in progress for CircleCI...

It looks like I should also add a Rust test to run the (non generated) C# unit tests.

matbd: @mattico Oh no, this is bad: CI didn't catch the two errors I introduced even if I revert my first commit

mattico: I can also confirm that C# tests aren't running on `master`. It appears the issue began on a5891983. It might be worthwhile to make the tests check that the test runner outputs `running <X> tests` where X > 0.

matbd: https://github.com/novifinancial/serde-reflection/pull/77

matbd:
Moved to new PR https://github.com/novifinancial/serde-reflection/pull/81

PR #80, opened 2020-12-21
Update binary serde

danprinz:
## Summary

- Update typescript runtime to use `util` for `TextEncoder`/`TextDecoder`
- Return null for Unit (instead of undefined)
- Handle more than double space needed for a new byte array insert (`ensureBufferWillHandleSize`)

## Test Plan

no change


matbd:
@danprinz Can you rebase this PR? (and/or squash some commits: the github UI shows that the resulting changes are very small)

matbd:
Thanks!

PR #84, opened 2021-02-19
[ci] simplify ci config

matbd:
## Summary

* remove apt-get upgrade
* use debian packages for npm and node-js instead of nvm
```
node v10.23.1
npm 5.8.0
```

## Test Plan

ci


JoelMarcey (approved):
LGTM 👍

Would you be open to seeing if we could run the CI on GitHub Actions?

matbd:
> LGTM 👍

Btw. we probably need to install (and use) typescript sometime.

> Would you be open to seeing if we could run the CI on GitHub Actions?

It depends if somebody is volunteering :) It could take a bit of work to setup the github page / documentation publishing action.

PR #89, opened 2021-03-08
[serde-generate] Fix --target-source-dir required message

mattico:
This argument is now named `--target-source-dir`


matbd:
Good catch, thanks!

PR #91, opened 2021-05-20
CSharp: Add Clone methods

mattico:
`MemberwiseClone` is a private method on `object` which performs a shallow clone. Classes can choose to expose this to their users, and we should.

We may want to implement a deep clone which would take some extra work. This could be exposed as a separate `DeepClone` method.


matbd:
@mattico Seems reasonable. Can you run `cargo fmt` to solve the lint error?

PR #92, opened 2021-05-24
Rust 1.52.1

matbd:
## Summary

* Prepare new release of serde-generate after https://github.com/novifinancial/serde-reflection/pull/91
* Update Rust toolchain

## Test Plan

CI


serde-generate/src/test_utils.rs
```diff
@@ -337,9 +337,9 @@ pub enum Runtime {
     Bincode,
 }
 
-impl std::convert::Into<Encoding> for Runtime {
-    fn into(self) -> Encoding {
-        match self {
+impl std::convert::From<Runtime> for Encoding {
```
JoelMarcey, line 340: Was this just wrong before?

JoelMarcey (approved):

serde-generate/src/test_utils.rs
matbd: No, implementing `std::convert::From<Bla> for Foo` is just better practice than implementing `std::convert::Into<Foo> for Bla` (as the latter can be deduced from the former)

PR #94, opened 2021-06-27
quote var names to eliminate javascript keywords

danprinz:
## Summary

In the latest code generation, a variable name called `function` was used. While you can use it in other languages, JavaScript forbids the use of keywords as variable names. 

## Test Plan

(Share your test plan here. If you changed code, please provide us with clear instructions for verifying that your changes work.)


matbd:
@danprinz Thanks for the fix. Would you be able to sign the CLA again?

matbd (approved):
If this happens often (and for other languages), I would suggest remapping the name(s) earlier by visiting the input formats (which can be done in language-generic way).

matbd:
I'm going to pause this idea for now. We're considering renaming things on the user side instead at the moment.

PR #95, opened 2021-07-05
Fixed typescript compile errors

Dinnerbone:
## Summary

The Typescript code doesn't compile right now, with the very fun and obscure error: `TS2345: Argument of type 'number | BigInt' is not assignable to parameter of type 'string | number | bigint | boolean'.`

`BigInt` is the name of the constructor, that takes a value and turns it into a `bigint`.

`bigint` is the type that is used to contain a big integer.

Technically I think it'd work (maybe without strict mode and less type checking), but Typescript wants the distinction made - and it matches the rest of the primitives (`number` vs `Number` etc).

This also fixes that the Typescript bincode serializer/deserializer doesn't compile because it changes the signature of `deserializeLen`/`serializeLen` to use bigint, which is incompatible with the parent signature of number.

In addition, I've made the constructor of binary de/serializer public so that you can actually make the subclasses. Making it protected and the subclasses override it to be public is no different than just making it public - save for that sometimes people forget to override it, as was the case here ;)

## Test Plan

- Try to run compile the existing code, Typescript will refuse with the above error.
- Try to run compile the new code, it works.

I've confirmed that it works when `install_serde_runtime()` is ran in my own code too.


Dinnerbone:
It looks like the build is failing but as far as I can tell, it's a network issue and not introduced by my code:

```
npm ERR! code E403
npm ERR! 403 Forbidden: @types/node@12.12.2

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2021-07-06T10_02_45_213Z-debug.log
test test_that_ts_code_compiles_with_bcs ... FAILED
npm ERR! code E403
npm ERR! 403 Forbidden: typescript@^3.9.6
npm
 ERR! npmcode  E403
ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2021-07-06T10_02_45_234Z-debug.log
npm ERR! 403 Forbidden: typescript@^3.9.6
test test_that_ts_code_compiles_with_comments ... FAILED

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2021-07-06T10_02_45_240Z-debug.log
test test_that_ts_code_compiles ... FAILED
```

matbd:
@danprinz @JoelMarcey Have you seen this issue before?

serde-generate/runtime/typescript/serde/binarySerializer.ts
```diff
@@ -17,7 +17,7 @@ export abstract class BinarySerializer implements Serializer {
     private buffer: ArrayBuffer;
     private offset: number;
 
-    protected constructor() {
```
matbd, line 20: Not a specialist of JS/TS but I suspect a protected constructor could make sense for an abstract class.

serde-generate/runtime/typescript/bincode/bincodeDeserializer.ts
```diff
@@ -1,8 +1,8 @@
 import {BinaryDeserializer} from "../serde/binaryDeserializer";
 
 export class BincodeDeserializer extends BinaryDeserializer {
-    deserializeLen(): BigInt {
```
matbd, line 4: I guess `Number` is indeed large enough for the maximal length authorized by the BCS specs (or supported by Java vectors anyway).

serde-generate/runtime/typescript/serde/binarySerializer.ts
```diff
@@ -84,7 +84,7 @@ export abstract class BinarySerializer implements Serializer {
         this.serializeWithFunction(DataView.prototype.setUint32, 4, value);
     }
 
-    public serializeU64(value: BigInt | number): void {
```
matbd, line 87: I don't understand this change. The JS community seems to be using `BigInt`.

serde-generate/runtime/typescript/serde/binarySerializer.ts
Dinnerbone: If the class is abstract the constructor cannot be called. Having a protected constructor only forces that the class overrides it just for the sake of making it not protected again, which introduces possible refactor pitfalls (if you add an argument to the constructor, remember to go change all the other constructors to mirror it - again, their only purpose is to set it back to public).

serde-generate/runtime/typescript/bincode/bincodeDeserializer.ts
Dinnerbone: It better be - it's being coerced back to `number` elsewhere! :D

serde-generate/runtime/typescript/serde/binarySerializer.ts
Dinnerbone: For the constructor. Every primitive in JS has a constructor and the type, which typically are only differentiated by capitalization. In JS world this doesn't matter, in Typescript it does - they are different types, one is a function and one is the actual value. This PR and the feature involved is about Typescript and not JS, and Typescript does not compile with this currently.

See that the code uses `number` and not `Number`, or `string` and not `String`. It's the same.

If you check the official documentation for BigInt in typescript, this will show that `bigint` is used as the type of a value that is big int, and `BigInt` is used as a function to convert an input into `bigint`. In Rust terms this is `Foo` vs `Foo::new()`.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#bigint

```typescript
let foo: bigint = BigInt(100); // the BigInt function
let bar: bigint = 100n; // a BigInt literal
```

Dinnerbone:
I can't test right now but I suspect a lot of this isn't being caught by your current tests because you may not be using strict mode. It's the recommended value and is default for most typescript project generators. I can add a commit to turn it on to prove these errors if you like.

matbd:
Thanks @Dinnerbone for the explanations. This all seems very reasonable.

> I can add a commit to turn it on to prove these errors if you like.

Yes please, that would be much appreciated.

matbd:
I'm going to merge this PR now. Thanks @Dinnerbone for the fixes!

PR #97, opened 2021-08-19
[serde-generate] Add support for nested byte vectors

tzakian:
## Summary

This PR adds support for serialization/deserialization of `vector<vector<u8>>` to the serde-generate runtimes for C#, Go, Java, and Python.

## Test Plan

Added additional tests to make sure serialization and deserialization works.

### DIFF
--- a/serde-generate/runtime/csharp/Serde/BinaryDeserializer.cs
+++ b/serde-generate/runtime/csharp/Serde/BinaryDeserializer.cs
@@ -65,6 +65,22 @@ namespace Serde
             return utf8.GetString(content);
         }
 
+        public ValueArray<ValueArray<byte>> deserialize_vec_bytes() {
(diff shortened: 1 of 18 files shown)


matbd:
@tzakian Why do we need this?

tzakian:
@matbd In the process of adding support for `vector<vector<u8>>` arguments to the transaction-builder generator I ran into issues where the underlying runtime for the generated code in C#, Go, Java, and Python didn't have support for serializing/deserializing vector<vector<u8>> (/Vector<Bytes>) and I couldn't see anywhere where support for that type existed, so this PR is adding in support for that type to the serialization/deserialization runtimes.

tzakian:
After discussion I'm going to close this PR out in favor of the newtype approach.

PR #101, opened 2021-09-15
Serde name adapter

matbd:
## Summary

Add support to override the Serde names of generic types using two "adapter" classes.

See README file for a motivational example:
https://github.com/matbd/serde-reflection/tree/serde_name_adapter/serde-name

## Test Plan

New unit tests in CI


serde-name/src/de_adapter.rs
```diff
@@ -0,0 +1,134 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use serde::de::Visitor;
+
+/// An adapter for `serde::de::Deserializer` implementations that lets you override the
+/// name of the top-level container.
+pub struct DeserializeNameAdapter<D> {
+    inner: D,
+    name: &'static str,
+}
```
bmwill, line 11: I guess the most surprising thing to me is that you're having to wrap the Deserializer/Serializer instead of the type you want to rename and this is due to the API of the Deserialize/Serialize traits not giving you this ability.

bmwill (approved):

serde-name/src/de_adapter.rs
matbd: Yes, credits go to @dtolnay for suggesting this approach.

PR #102, opened 2021-09-15
[serde-reflection/generate] improve documentation

matbd:
## Summary

Improve documentation of serde-reflection and serde-generate to address common questions.

## Test Plan

CI

### DIFF
--- a/serde-generate/Cargo.toml
+++ b/serde-generate/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.19.3"
+version = "0.19.4"
 description = "Generate (de)serialization code in multiple languages"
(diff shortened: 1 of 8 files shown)


serde-generate/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "serde-generate"
-version = "0.19.3"
+version = "0.19.2"
```
bmwill, line 3: Did you mean to drop the version?

matbd: oops. good catch

matbd: I suppose it passes CI because minor versions can be silently promoted.

bmwill (approved):

PR #105, opened 2021-10-05
[serde-generate] update CI and docs for typescript

matbd:
## Summary

Follow up with #95 

## Test Plan

CI

### DIFF
--- a/serde-generate/tests/typescript_generation.rs
+++ b/serde-generate/tests/typescript_generation.rs
@@ -47,8 +47,7 @@ fn write_package_tsconfig_json_for_test_build(path: std::path::PathBuf) -> Resul
     "forceConsistentCasingInFileNames": true,
     "lib": [
       "es6",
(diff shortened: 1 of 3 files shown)


serde-generate/tests/typescript_generation.rs
```diff
@@ -46,9 +46,8 @@ fn write_package_tsconfig_json_for_test_build(path: std::path::PathBuf) -> Resul
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "lib": [
-      "es6",
-      "esnext.BigInt",
-      "dom"
+//      "es6",
```
dimroc, line 49: In L40 above, we target es6. With it no longer being esnext, perhaps BigInt is not in the standard without the need for libs?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects#numbers_and_dates

PR #106, opened 2021-10-06
Replace node runtime with deno for typescript

dimroc:
## Summary

Replace node support with [deno](https://deno.land/). Advantages of deno:
1. Easier setup and package management, less hassle with tsconfig and npm
2. More secure defaults
3. Bundles to JS to better support browser scenarios

Node can still be supported with either the bundle-to-js step above, or simple change of `import utils...` line.

Requires installation of deno:

```
brew install deno
```

## Test Plan

cargo test


serde-generate/Cargo.toml
```diff
@@ -19,6 +19,7 @@ exclude = [
 heck = "0.3.2"
 include_dir = "0.6.0"
 maplit = "1.0.2"
+regex = "1.5.4"
```
matbd, line 22: Please move to `dev-dependencies`.

serde-generate/src/common.rs
```diff
@@ -41,3 +42,57 @@ pub(crate) fn mangle_type(format: &Format) -> String {
         Variable(_) => panic!("unexpected value"),
     }
 }
+
+#[derive(Debug)]
+pub enum TypeScriptRuntime {
```
matbd, line 47: How about having this code simply in `typescript.rs` ?

serde-generate/src/deno.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Facebook, Inc. and its affiliates
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::{common::TypeScriptRuntime, typescript, CodeGeneratorConfig};
```
matbd, line 4: Let's discuss a bit more how to structure the modules and share the code here.

serde-generate/src/typescript.rs
```diff
@@ -88,24 +87,24 @@ where
...
+"#,
+        );
         for namespace in self.generator.namespaces_to_import.iter() {
-            writeln!(
-                self.out,
-                "import * as {} from '../{}';\n",
-                namespace.to_camel_case(),
-                namespace
-            )?;
+            rval.push_str(
```
matbd, line 98: Looks like we could bring back `writeln!` and avoid string concatenation

serde-generate/tests/typescript_runtime.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::{fs::File, io::Write, process::Command};
+use tempfile::tempdir;
+
+#[test]
+fn test_typescript_bcs_runtime_on_simple_data() {
+    test_typescript_runtime_on_simple_data(Runtime::Bcs);
+}
+
+fn test_typescript_runtime_on_simple_data(runtime: Runtime) {
+    let registry = test_utils::get_simple_registry().unwrap();
```
matbd, line 19: Thanks for adding this test.

serde-generate/tests/typescript_generation.rs
```diff
@@ -1,127 +1,91 @@
...
+    ];
+
+    for (input, expectation) in table {
+        assert_eq!(is_error_output(input), expectation);
+    }
 }
 
 #[test]
-fn test_that_ts_code_compiles_with_bcs() {
+fn test_typescript_code_compiles_with_bcs() {
```
matbd, line 73: I'm assuming this includes type-checking?

matbd: ok it does

PR #112, opened 2021-10-20
Add support for Dart 2.14+

jerel:
## Summary

This PR brings #57 up to date with the latest code on main and also adds a number of fixes and improvements. As we have a project based on serde-reflection we have motivation to keep the Dart implementation current.

Generated Dart code follows the official [Effective Dart](https://dart.dev/guides/language/effective-dart) guide and CI requires generated code to pass `dart analyze` as well as `dart test`.

## Test Plan
The following usage should produce a working Dart project:
``` rust
let registry = test_utils::get_registry().unwrap();

let config = CodeGeneratorConfig::new(String::from("example"))
    .with_encodings(vec![Encoding::Bincode])
    .with_c_style_enums(true);

let package = PathBuf::from("dart_example");
let installer = dart::Installer::new(package.clone());
installer.install_bincode_runtime().unwrap();
installer.install_bcs_runtime().unwrap();
installer.install_serde_runtime().unwrap();
installer.install_module(&config, &registry).unwrap();
```


serde-generate/src/dart.rs
```diff
@@ -0,0 +1,1283 @@
...
+            match self.get_field_container_type(name) {
+                Some(Enum(_)) => format!("{}Extension", name),
+                _ => name.to_string(),
+            }
+        } else {
+            name.to_string()
+        }
+    }
+
+    fn quote_qualified_name(&self, name: &str) -> String {
```
matbd, line 257: So until now, we have not attempted to remap conflicting names on the fly with a list like this in each language. (I feel like the list may always be growing)

If we must, perhaps we can make this behavior a codegen configuration option with a default value that is documented for each language.

@dimroc what do you think?

serde-generate/src/dart.rs
```diff
@@ -0,0 +1,1283 @@
...
+                }
+            }
+        }
+
+        // Fields
+        if !fields.is_empty() {
+            writeln!(self.out)?;
+        }
+        for field in fields {
+            //self.output_comment(&field.name)?;
```
matbd, line 756: Did you mean to remove or uncomment this later?

serde-generate/tests/dart_runtime.rs
```diff
@@ -0,0 +1,38 @@
...
+fn install_test_dependency(path: PathBuf) -> Result<()> {
+    Command::new("dart")
+        .current_dir(path)
+        .args(["pub", "add", "-d", "test"])
+        .status()?;
+
+    Ok(())
+}
+#[test]
+fn test_dart_runtime() {
```
matbd, line 14: So we are quite testing interoperability of serialization formats here. That's fine but let's document the language as "in progress".

serde-generate/README.md
```diff
@@ -16,6 +16,7 @@ It can be used as a library or as a command-line tool (see `serdegen` below).
 The following programming languages are fully supported as target languages:
 
 * C++ 17
+* Dart >= 2.14
```
matbd, line 19: It's probably a bit early to claim that Dart is fully supported (see comments on unit tests). Do you mind moving this to the next section with TypeScript 4 ?

serde-generate/runtime/dart/serde/binary_serializer.dart
```diff
@@ -0,0 +1,134 @@
...
+    output.addAll(bdata.buffer.asUint8List());
+  }
+
+  void serializeUint64(Uint64 val) {
+    BigInt number = val.toBigInt();
+    final _byteMask = BigInt.from(0xFF);
+    int bytes = 8;
+    var bdata = Uint8List(bytes);
+    for (int i = 0; i < bytes; i++) {
+      // big endian
```
matbd, line 57: I would rather remove commented code here and in general. (Hopefully you don't need big endian, do you?)

serde-generate/runtime/dart/test/bcs.dart
```diff
@@ -0,0 +1,43 @@
+import 'dart:typed_data';
```
matbd, line 1: Please make a pass and add the missing copyright headers.

matbd (commented):
Thank you for the great PR! I left a few comments. This looks good overall and I feel that we should be able to land it on master soon, if we are clear that this language support is sill stabilizing.

matbd:
Not sure why we don't see the result of the CI (which we definitely need). Perhaps try to rebase?

serde-generate/tests/dart_generation.rs
```diff
@@ -86,15 +89,20 @@ fn test_dart_code_compiles_with_comments() {
...
+    let tempdir = tempdir().unwrap();
+    let source_path = tempdir.path().join("dart_project");
 
     let config = CodeGeneratorConfig::new("example".to_string())
         .with_encodings(vec![Encoding::Bcs, Encoding::Bincode])
         .with_c_style_enums(false);
 
     test_dart_code_compiles_with_config(source_path, &config);
+
+    tempdir.close().unwrap();
```
matbd, line 107: You should not be needing that if `tempdir()` is the right thing.

matbd: which it is

https://docs.rs/tempfile/3.2.0/tempfile/fn.tempdir.html

PR #109, opened 2021-10-12
Adding swift support

texas16:
## Summary

Initial Swift (iOS) support.

## Test Plan

Unit tests will be added.

PR #113, opened 2021-10-22
Swift code-generation

matbd:
## Summary

Extends https://github.com/novifinancial/serde-reflection/pull/109 :
* CI under Linux
* code generation module should be mostly done
* several fixes in the runtime

Caveats and TODOs:
* It looks like the "Reader" runtime still crashes ("illegal instruction 4") in case of underflow.
* BCS maps are not yet supported at runtime
* Not sure how to solve the warning `BinaryReader.swift:83:63: note: condition always evaluates to true`. Perhaps we should give up on testing endianness (which itself does not seem very portable) and just use the simpler (mildly slower) arithmetics as in other languages.
* We still need to port the more complicated runtime tests to confirm that serialization and deserialization work as intended for all supported types.
* I personally gave up on using Swift tuples (not equatable or hashable)
* The current generated code marks everything as "indirect". We could probably do much better in the future but this should not be blocking.

## Test Plan

CI

Manual testing with Diem types:
```
wget https://raw.githubusercontent.com/diem/diem/main/testsuite/generate-format/tests/staged/diem.yaml

cargo run --bin serdegen -- --with-runtimes serde bcs --module-name Diem --language swift diem.yaml --target-source-dir test

cat >test/Package.swift <<EOF
// swift-tools-version:5.3

import PackageDescription

let package = Package(
    name: "Diem",
    products: [
        .library(
            name: "Diem",
            targets: ["Diem"]
        ),
    ],
    targets: [
        .target(
            name: "Serde",
            dependencies: []
        ),
        .target(
            name: "Diem",
            dependencies: ["Serde"]
        ),
    ]
)
EOF

(cd test && swift build)
```


matbd:
Also, there is this suspicious warning:
```
serde-generate/runtime/swift/Sources/Serde/BigInt.swift:130:24: warning: function call causes an infinite recursion
                self = _BigInt(sourceMag)
```

PR #116, opened 2021-11-04
Dart tests

jerel:
## Summary

This is a follow-up to #112 to adjust tests.


matbd (approved):

matbd:
@jerel Just to be sure: Do you mean to land this PR now or continue extending tests here?

jerel:
@matbd I think I'll keep pushing more test updates to this branch, that way it gives me a few commits to watch the CI stability

jerel:
@matbd this is ready for another look. This PR now:
* Makes a local pub-cache (in the `tempdir`) for Dart dependencies to avoid any dependency install/uninstall conflict between test cases.
* Always copies the serde test in but conditionally brings in `bcs.dart` and `bincode.dart` tests if the project uses that encoding.
* Adds a `runtime.dart` test file that is only brought in by the `dart_runtime.rs` tests. Those tests test against the `test_utils.rs` data structures with both Bincode and BCS encodings and with both flavors of enums.

serde-generate/tests/dart_runtime.rs
```diff
@@ -14,13 +15,56 @@ fn install_test_dependency(path: &Path) -> Result<()> {
 
     Ok(())
 }
+
 #[test]
 fn test_dart_runtime() {
     let tempdir = tempdir().unwrap();
     let source_path = tempdir.path().join("dart_project");
 
     let registry = test_utils::get_registry().unwrap();
```
matbd, line 24: The way I understand this test it seems to be superficially using the Rust module `test_utils` to generate some format then import (templated) Dart code from `runtime/dart/test/` and verify that deserialization work in the generated code.

The problem that I have with this approach is that

1. It is completely different from what we have in other languages so IMHO it's creating more maintenance burden than needed just for Dart

2. What happens if I want to change `get_registry()` later? Currently, this impacts only tests written in Rust with quoted code in other languages, and most of the format-dependent logics is shared between languages. Here we have to maintain tight consistency with a Dart templated file in `runtime/dart`.

3. This approach does not test interoperability with Rust bytes as the other xxx_runtime.rs files do

4. It cannot be easily extended to test the many corner cases of (de)serialization as we did in other languages.

matbd (changes requested):
Thanks for the PR. Let's discuss more about the approach here (see comments above, let me know if they make sense)

serde-generate/tests/dart_runtime.rs
jerel: @matbd thanks for the feedback, that makes sense. I had somehow missed (!) that the `reference.map()` code in other runtime tests was injecting the Rust bytes into the runtime code and so I was focused on just "testing the runtime". I've pushed up an implementation that's almost identical to Typescript/Go/etc now if you want to take a look.

serde-generate/runtime/dart/test/runtime.dart
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Facebook, Inc. and its affiliates
```
matbd, line 1: Do you still need this file?

serde-generate/src/dart.rs
```diff
@@ -1218,7 +1218,21 @@ impl crate::SourceInstaller for Installer {
         let generator = CodeGenerator::new(config);
         generator.output(self.install_dir.clone(), registry)?;
         generator.output_test(&self.install_dir)?;
-        self.install_runtime(include_directory!("runtime/dart/test"), "test/src")?;
+        std::fs::create_dir_all(self.install_dir.join("test/src")).unwrap();
+
+        let tests = include_directory!("runtime/dart/test");
```
matbd, line 1223: This part still feels a bit complicated. Perhaps we can just not copy the unit tests of the runtime library (or copy them as part of the install_*_runtime methods ?)

matbd:
So I have an example of sporadic issue with Dart in CI :(

https://app.circleci.com/pipelines/github/novifinancial/serde-reflection/723/workflows/7514dbb4-274c-4a1f-8503-3cb219dc0f25/jobs/2151/parallel-runs/0/steps/0-105

or equivalently: https://gist.github.com/matbd/2f902f5d6f4a9ef08c1d60a05446cad3

It looks like that's maybe what you're fixing here with the change on `.pub-cache`? (I might just apply it right away then)

jerel:
@matbd yes, the `.pub-cache` change is an attempt at fixing that intermittent failure. Once that is in there shouldn't be any overlap at all between tests so I expect that to fix it. At least I was able to recreate that error locally by tampering with the `.pub-cache` between tests.

serde-generate/src/dart.rs
jerel: I'm happy to defer to your preference on what we should do with unit tests. In a way it feels unnecessary to me to install tests into the end user's project. But there was logic like this present previously so I kept it and just modified to only install the enabled encodings. I'm working on this today so can pull it out right now if you like.

jerel: Looking at some of the other implementations I think my pick would be to get rid of Dart's `output_test` and drop this test copying logic from `install_module`, then copy the individual files in the `dart_runtime.rs` tests. Thoughts?

jerel: Here's what that would look like https://github.com/jerel/serde-reflection/commit/d46f6bdde9b0b269afc6a1567cec1b0f854975a7 If you like that approach I can add it to this PR.

matbd: Yes please let's not generate or copy test during codegen. We do run unit tests as part of the `xx_runtime.rs` tests files under a separate test often called `xx_autotest` or equivalent

jerel: Ok, the above commit is added to the PR but as `test_dart_runtime_autotest` instead of running with the encoding runtime tests. I'm finished with this PR unless there's something else that you see.

matbd (approved):
Alright I have one last question but let's land this PR first then I'll explain

matbd:
@jerel So my question is what's the problem with this approach for the autotest then: https://github.com/novifinancial/serde-reflection/pull/118

PR #117, opened 2021-11-11
[Swift] Finalize runtime and add extended runtime tests

matbd:
## Summary

* Generally attempt to simplify the Swift runtime (without thinking too much about performance) and harden everything
     * Avoid CFFoundation (e.g. OutputStream.toMemory() didn't seem to be an extensible buffer) and use the copy-on-write Swift arrays as buffers everywhere we can
     * Avoid pulling big chunk of code for 128-bit integers (not used in Diem anyway)
     * Try to consistently mark `throws` public APIs that could throw in the future
     * Review all integer conversions: by default conversion underflows/overflows crash at runtime which is really nasty
* Makes booleans and maps more strict in BCS (covered by the Rust-driven fuzzing tests)
* Set BCS limits on sequences (same)

## Test Plan

new runtime tests in CI


tzakian (approved):
LGTM! Really like the consistency around `throw` on the deserialization functions :)

serde-generate/runtime/swift/Sources/Serde/BinaryDeserializer.swift
```diff
@@ -51,100 +59,108 @@ public class BinaryDeserializer: Deserializer {
...
     }
 
-    public func deserialize_u128() throws -> BigInt8 {
-        let signed: BigInt8 = try deserialize_i128()
-        if signed >= 0 {
-            return signed
-        } else {
-            return signed + (BigInt8(1) << 128)
-        }
+    public func deserialize_u128() throws -> UInt128 {
```
texas16, line 122: does this method throws?

PR #118, opened 2021-11-15
[dart] make `dart test` work from runtime/dart

matbd:
## Summary

This is me trying to make the basic Dart unit tests runnable from the CLI.

## Test Plan

```
cd serde-generate/runtime/dart
dart test
```

### DIFF
--- a/serde-generate/README.md
+++ b/serde-generate/README.md
@@ -26,6 +26,7 @@ The following programming languages are fully supported as target languages:
 The following languages are partially supported and still considered under development:
 
 * TypeScript 4 (packaged and tested with Deno)
+* Dart >= 2
 
 ### Supported Encodings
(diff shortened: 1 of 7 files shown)


jerel:
I wasn't sure if the `runtime/*` folders were meant to be runnable or just source directories but if you're cool with the pubspec being added in there it seems good to me. Simpler than doing the copy and makes it easier/faster to add new tests.

matbd:
> I wasn't sure if the `runtime/*` folders were meant to be runnable or just source directories but if you're cool with the pubspec being added in there it seems good to me. Simpler than doing the copy and makes it easier/faster to add new tests.

Thanks for the comment. Great! Yes, we do have a few build/package files for some languages.

PR #122, opened 2021-12-03
[serde-generate] fix empty struct/tuple variants in python and swift

matbd:
## Summary

Support `enum Foo { Bar(), Baz{} }`

## Test Plan

new test cases + CI


tzakian (approved):
Thanks for this!

PR #125, opened 2022-04-05
Dart utf8

jerel:
## Summary

Initial implementation of the Dart runtime handled string ser/de with [utf-16] code units. This had the appearance of working as long as the given character was < 128. This PR switches to utf-8 encode/decode to fix this bug.

## Test Plan

A test is included with Japanese text which causes the current implementation to fail.

### DIFF
--- a/serde-generate/runtime/dart/serde/binary_deserializer.dart
+++ b/serde-generate/runtime/dart/serde/binary_deserializer.dart
@@ -111,7 +111,7 @@ abstract class BinaryDeserializer {
   int deserializeVariantIndex();
 
   String deserializeString() {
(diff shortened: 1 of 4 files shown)


ma2bd:
Thanks! FYI we're in the process of moving this repo (I can't accept PRs here)

jerel:
Ok, thanks for the update! Let me know if there's anything I can do such as opening against a different repo, etc.

ma2bd:
@jerel Hi! Still no progress on an official repo migration. Can you make a PR on github.com/zefchain/serde-reflection ?

jerel:
Yes, happy to do that.

And... done.

